PR #14, opened 2026-02-24
feat: Connections page domain management

bderenzi:
## Summary

- Replaces the broken Connections page (which only showed OAuth providers and went blank when none were configured) with a full domain management UI
- Add, edit, and remove CommCare API-key domains inline, alongside the existing OAuth providers section
- Adds a `PATCH /api/auth/tenant-credentials/<id>/` endpoint for updating credentials

## Changes

**Backend**
- New `PATCH` endpoint on `tenant_credential_detail_view` — updates `tenant_name` and/or `encrypted_credential` (partial update supported)
- New tests: PATCH updates credential, requires auth, 404 for wrong user, DELETE removes domain from tenants list

**Frontend**
- `ConnectionsPage` rewritten with two sections: CommCare Domains (API Key) and OAuth Providers
- Inline edit: clicking Edit replaces the card in-place with the edit form
- Inline remove confirmation: clicking Remove flips the card to a confirmation row ("Confirm Remove") — no `window.confirm()`
- Remove button styled with soft red (light background, dark red border/text)
- After add/remove, both local domain list and the sidebar Zustand store are refreshed
- If the removed domain was the active selection in the sidebar, automatically switches to the next available domain

## Test plan
- [ ] Add a domain — appears in list and sidebar dropdown
- [ ] Edit a domain name — updates in place
- [ ] Remove a domain — confirmation row appears, confirm removes it from list and dropdown
- [ ] Remove the currently-active domain — dropdown switches to next available
- [ ] All backend tests pass: `uv run pytest tests/test_tenant_api.py tests/test_auth.py`


bderenzi:
Note for @snopoke: the `lint-python` and `test` failures in earlier CI runs were pre-existing on `main` — not introduced by this PR. Specifically:

- `scripts/setup_oauth_apps.py` had an unsorted import block (I001) — fixed in the latest commit
- `test_tool_content_to_str_handles_types` is a pre-existing failing test unrelated to these changes

Also just got pre-commit hooks installed locally so ruff will catch issues before they hit CI going forward.

PR #54, opened 2026-03-04
feat: add /download-commcare-knowledge slash command and tool

jjackson:
## Summary
- Adds `/download-commcare-knowledge` slash command that fetches CommCare app form definitions from HQ and saves them as KnowledgeEntry records
- Creates a new LangGraph tool (`download_commcare_knowledge`) that the agent invokes to fetch learn/deliver app structures for Connect opportunities
- Extracts `get_commcare_credential` into a shared utility (`apps/agents/utils/commcare_auth.py`) for reuse
- Makes `fetch_app_by_id` and `extract_form_definitions` public in `commcare_metadata.py`

## Motivation
The agent needs to know what fields exist in `visits.form_json`. Previously we tried annotating this during materialization, but that couples data sync with knowledge. This approach makes knowledge an explicit, user-triggered step via a slash command.

## Test plan
- [x] 8 new tests covering content building, error paths, create, and update flows
- [x] `uv run pytest tests/test_commcare_knowledge_tool.py -v` — all passing
- [x] Manual: verified tool is invoked by agent and returns expected results


snopoke (commented):
This overlaps with what we're already pulling from CC here: https://github.com/dimagi-rad/scout/blob/main/mcp_server/loaders/commcare_metadata.py#L92.

This get's used during the 'describe_table' tool call though we are barely using it at the moment: https://github.com/dimagi-rad/scout/blob/main/mcp_server/services/metadata.py#L150

We still haven't implemented any transformations from the raw data. One place where this metadata could be used is in that transformation process to 'explode' form data into more queryable formats.

@bderenzi do you have a sense of the best way to expose this?

PR #126, opened 2026-04-07
docs: OCS integration design spec

theism:
The same way Scout will be helpful to analyze Connect or CommCare data, it will be very helpful to be able to leverage it for OCS data as well.  This design doc is Matt's attempt to feed Claude the OCS docs, API information, and existing implementation to add  Open Chat Studio as a third data source in Scout. It covers OAuth provider, tenant resolution, pipeline, loaders, and materialization — following the existing Connect provider pattern.

The Connect team is creating a number of ways to analyze OCS data via AI and Scout replace/ease the ability for any of the teams leveraging OCS bots to do that, namely and most immediately, the Interview team.

### DIFF
--- a/docs/superpowers/specs/2026-04-06-ocs-integration-design.md
+++ b/docs/superpowers/specs/2026-04-06-ocs-integration-design.md
@@ -1,0 +1,216 @@
+# OCS (Open Chat Studio) Integration Design
+
+**Date:** 2026-04-06
+**Status:** Draft
+
+## Overview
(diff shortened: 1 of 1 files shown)


jjackson:
@snopoke I'm taking a run at this, for hygiene, should we record that we talked about how this looked mostly good and/or have states that these are in so we know they have gone through the design review process?

jjackson:
implemented in #152

PR #136, opened 2026-04-09
Fix DB connection reliability and schema provisioning

bderenzi:
## Summary
- **autocommit fix**: Use `autocommit=True` on MCP query connections to prevent "current transaction is aborted" cascade errors. When `SET ROLE` or a query fails inside an implicit transaction, all subsequent statements fail with a confusing error. Autocommit isolates each statement.
- **slugify fix**: Handle CommCare form definitions where `name` is a localized dict (`{"en": "Household Registration"}`) instead of a string. Extracts the English name, falls back to first available language, then xmlns.
- **provision fix**: `SchemaManager.provision()` now ensures the physical PostgreSQL schema and readonly role exist even when a Django `TenantSchema` record already says ACTIVE. Previously, if the physical schema was dropped externally, provision returned the stale record without recreating it.

## Test plan
- [x] All 25 query tests pass
- [x] All 50 staging/slugify tests pass
- [x] All 10 schema manager tests pass (updated test for new behavior)
- [x] Deployed and verified queries work end-to-end on production

### DIFF
--- a/apps/transformations/services/commcare_staging.py
+++ b/apps/transformations/services/commcare_staging.py
@@ -288,6 +288,9 @@ def generate_system_assets(tenant, metadata: dict) -> list[TransformationAsset]:
 
     for xmlns, form_def in form_definitions.items():
         form_name = form_def.get("name", xmlns)
+        if isinstance(form_name, dict):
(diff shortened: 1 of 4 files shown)


bderenzi:
Already deployed to production. PR for review record only — merging.

PR #141, opened 2026-04-13
Multi-tenant workspace materialization

snopoke:
## Summary

- Refactor `run_materialization` MCP tool to operate at the workspace level — resolves all tenants in a workspace and runs the appropriate pipeline for each, replacing the single-tenant `tenant_id` interface
- Fix view schema permissions: grant SELECT on views explicitly since `ALTER DEFAULT PRIVILEGES` only covers future tables, not the views created during materialization
- Fix hardcoded `sslmode=require` in MCP context — now reads `sslmode` from the database URL query string so local dev (no SSL) works correctly
- Add Celery worker with auto-reload to dev Procfile

## Test plan

- [ ] Run materialization on a workspace with multiple tenants and verify all tenant schemas are materialized
- [ ] Verify view schema views are queryable (SELECT on views was previously broken)
- [ ] Test with database URLs with and without `sslmode` parameter
- [ ] Start dev environment with `honcho -f Procfile.dev start` and confirm Celery worker starts and auto-reloads

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/mcp_server/server.py
+++ b/mcp_server/server.py
@@ -495,165 +495,192 @@ def _resolve_oauth_credential(token_obj, provider: str) -> dict:
     return {"type": "oauth", "value": token_value}
 
 
+async def _materialize_tenant(
+    tm,
+    pipeline_config,
+    ctx: Context | None,
+) -> dict:
+    """Run a materialization pipeline for a single TenantMembership.
+
+    Resolves credentials, builds a progress callback, and executes the pipeline.
+    Returns the pipeline result dict on success, or raises on failure.
+    """
+    from apps.users.models import TenantCredential
+    from mcp_server.loaders.commcare_base import CommCareAuthError
+    from mcp_server.loaders.connect_base import ConnectAuthError
+
+    tenant_id = tm.tenant.external_id
+
+    # ── Resolve credential ────────────────────────────────────────────────
+    try:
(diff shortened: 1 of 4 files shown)


mcp_server/server.py
```diff
@@ -495,165 +495,192 @@ def _resolve_oauth_credential(token_obj, provider: str) -> dict:
...
+        credential = {"type": "api_key", "value": decrypted}
+    else:
+        from allauth.socialaccount.models import SocialToken
+
+        if tm.tenant.provider == "commcare_connect":
+            token_obj = (
+                await SocialToken.objects.select_related("app")
+                .filter(
+                    account__user=tm.user,
+                    account__provider__startswith="commcare_connect",
```
snopoke, line 539:
```suggestion
                    account__provider="commcare_connect",
```

mcp_server/server.py
```diff
@@ -495,165 +495,192 @@ def _resolve_oauth_credential(token_obj, provider: str) -> dict:
...
+                .afirst()
+            )
+        else:
+            token_obj = (
+                await SocialToken.objects.select_related("app")
+                .filter(
+                    account__user=tm.user,
+                    account__provider__startswith="commcare",
+                )
+                .exclude(account__provider__startswith="commcare_connect")
```
snopoke, line 550:
```suggestion
                    account__provider="commcare",
                )
```

PR #177, opened 2026-05-25
feat(auth): restrict OAuth logins to allow-listed email domains

snopoke:
## Summary

Adds a per-OAuth-provider email-domain allow-list, enforced in the existing `EncryptingSocialAccountAdapter` via `pre_social_login`. Default behavior restricts the three Dimagi-internal providers (`commcare`, `commcare_connect`, `ocs`) to `@dimagi.com` emails; deployments can override via the `SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS` env var (JSON object keyed by allauth provider id).

Blocked users are redirected to `/accounts/login/` with a flash message and never get a `User`, `SocialAccount`, or tenant-resolution side effects. Providers absent from the dict (or mapped to an empty list) are unrestricted — so `google` and `github` are not subject to this policy. OAuth responses that don't carry an email are allowed through (covers CommCare Connect, which doesn't return one).

## Design

- Spec: `docs/superpowers/specs/2026-05-25-oauth-domain-restriction-design.md`
- Plan: `docs/superpowers/plans/2026-05-25-oauth-domain-restriction.md`

## Behavioral notes for reviewers

- Enforcement applies to every OAuth login on the three covered providers — new signups AND existing users re-authenticating. Existing accounts whose email isn't on the allow-list will be locked out of OAuth login on next attempt (they can still password-login if a password was set).
- `SOCIALACCOUNT_EMAIL_REQUIRED = False` (pre-existing) interacts with this feature: providers that return no email pass the check silently. This is intentional — Connect doesn't return an email.
- Domain match is case-insensitive but exact — `user@sub.dimagi.com` is rejected when the allow-list is `["dimagi.com"]`. Test pins this behavior.

## Test plan

- [x] `uv run pytest tests/test_oauth_domain_restriction.py` — 11 tests, all pass
- [x] `uv run pytest tests/test_oauth_tokens.py` — 19 existing tests, no regression
- [ ] Manual smoke: log in via CommCare HQ with a @dimagi.com account on staging → allowed
- [ ] Manual smoke: log in via CommCare HQ with a non-dimagi account on staging → redirected to login with flash message

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -228,6 +228,24 @@ SOCIALACCOUNT_PROVIDERS = {
     },
 }
 
+# OAuth email-domain restriction.
+# Map of allauth provider id -> list of allowed email domains (lowercase, exact match).
+# A provider absent from the dict (or mapped to an empty list) is unrestricted.
+# A provider with a non-empty list rejects OAuth logins whose email domain isn't in the list.
+# A login that returns no email is allowed regardless (best-effort: covers providers like
+# Connect that don't return an email).
+# Override at deploy time with the SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS env var (JSON).
+SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS: dict[str, list[str]] = env.json(
+    "SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS",
+    default={
+        "google": ["dimagi.com"],
+        "github": ["dimagi.com"],
+        "commcare": ["dimagi.com"],
+        "commcare_connect": ["dimagi.com"],
+        "ocs": ["dimagi.com"],
+    },
+)
+
 
 # Django REST Framework settings
 REST_FRAMEWORK = {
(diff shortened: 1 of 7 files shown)


config/settings/base.py
```diff
@@ -228,6 +228,24 @@
...
+# A provider absent from the dict (or mapped to an empty list) is unrestricted.
+# A provider with a non-empty list rejects OAuth logins whose email domain isn't in the list.
+# A login that returns no email is allowed regardless (best-effort: covers providers like
+# Connect that don't return an email).
+# Override at deploy time with the SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS env var (JSON).
+SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS: dict[str, list[str]] = env.json(
+    "SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS",
+    default={
+        "google": ["dimagi.com"],
+        "github": ["dimagi.com"],
```
snopoke:
```suggestion
```

tests/test_oauth_domain_restriction.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class TestAllowedEmailDomainsSetting:
+    """Verify the SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS setting shape and defaults."""
+
+    def test_setting_exists_and_is_dict(self):
+        assert isinstance(settings.SOCIALACCOUNT_ALLOWED_EMAIL_DOMAINS, dict)
+
+    def test_default_restricts_all_five_providers_to_dimagi_com(self):
+        expected_providers = {"google", "github", "commcare", "commcare_connect", "ocs"}
```
snopoke:
```suggestion
        expected_providers = {"commcare", "commcare_connect", "ocs"}
```

PR #212, opened 2026-05-29
Redesign workspace switcher for hundreds of workspaces with a live data indicator

bderenzi:
## Summary

Redesigns the workspace switcher for accounts with **hundreds of workspaces** (typical: ~135 CommCare Connect + ~139 OCS = ~270). The old switcher was a `w-56` popover with a search box and a single scrollable list of *all* workspaces — a tiny painful scrollbar, no recents, and no honest signal for which workspaces actually have data.

The new `WorkspaceSwitcher` (`frontend/src/components/WorkspaceSwitcher/`), extracted out of `Sidebar.tsx`:

- **Recents-first.** Opens to a Recent section (up to 8, newest-first), tracked client-side in localStorage (`lib/recentWorkspaces.ts`) since the backend has no "recent" concept. Recents are snapshotted on open so selecting one doesn't reshuffle the list mid-use.
- **Search** across workspace display names *and* tenant names.
- **Provider segment pills** (`Recent / All / CommCare Connect / OCS / …`, with counts) that **wrap** instead of scrolling sideways — nothing is ever clipped in the 352px popover. Pills only render for accounts large/varied enough to need them (>1 provider or ≥12 workspaces), so small accounts stay uncluttered.
- **Compact "Has data" toggle** — a round emerald-dot icon button pinned top-right so it never competes with the pills for width.
- **Row windowing** (hand-rolled, no new dependency) keeps render cost flat at ~270 items.
- **Keyboard nav**: ↑/↓ move the highlight, Enter selects the highlighted (or top) match, Esc closes.

### Live, three-state data indicator (changed semantics)

The previous dot used `last_synced_at != null` — a **historical** "was synced at least once" signal that never disappeared when a workspace's data was torn down, so it could lie about current availability.

The per-row indicator is now driven by the **live** `schema_status` and has three states:

| State    | `schema_status` | Visual                      | Tooltip / aria             |
|----------|-----------------|-----------------------------|----------------------------|
| loading  | `provisioning`  | spinner (`animate-spin`)    | "Loading data…"            |
| ready    | `available`     | emerald dot                 | "Has data — synced \<ago\>" |
| empty    | `unavailable`   | hollow muted dot            | "No data"                  |

When data is torn down the row correctly returns to the hollow "No data" dot. The new frontend helper `workspaceDataState(ws)` derives this, falling back to `last_synced_at` only when `schema_status` is absent (older cached payloads), so the UI degrades safely. `workspaceHasData` now means "ready" and backs the "Has data" filter.

### Backend

- `schema_status` is now also returned by the workspace **LIST** endpoint (previously detail-only). `last_synced_at` is still returned for the "synced X ago" tooltip.
- The computation is factored into shared helpers in `apps/workspaces/api/workspace_views.py`: `_derive_schema_status(...)` (pure logic, used by both detail and list, so they never drift) and `_schema_status_for_workspaces(...)` which computes status for many workspaces with **two bulk queries** (per-tenant `TenantSchema` states + multi-tenant `WorkspaceViewSchema` states) — **no N+1**.

No data-model rename or migration — this uses existing schema semantics.

### Housekeeping

- Removed the scratch `SWITCHER-design.md` (its rationale lives here now).
- Restored the `.playwright*/` `.gitignore` entries that had been inadvertently dropped on the branch.

## Test plan

- `tests/test_workspace_last_synced.py` — 12 tests green, including new cases asserting list `schema_status` is `unavailable`/`available`/`provisioning` as expected and that **list `schema_status` matches detail** (no drift).
- `cd frontend && bun run lint` — clean.
- `cd frontend && bun run build` (tsc + vite) — clean.
- `uv run --extra dev ruff check .` — clean.
- Manual: open the switcher → recents shown first; pills wrap with no horizontal scroll; "Has data" toggle filters to ready workspaces; indicator shows spinner while provisioning, emerald dot when ready, hollow when empty; keyboard ↑/↓/Enter/Esc work.


snopoke:
Nice. One thing still missing is a quick link to 'manage this workspace' - you still have to go to 'manage workspaces' and filter that list to find the one.

PR #220, opened 2026-06-05
Multi-team OCS connections (first-class TenantConnection)

bderenzi:
## Multi-team OCS connections

Lets a user work with **multiple Open Chat Studio (OCS) teams** and reliably materialize each chatbot with the correct team's credential. Fixes the 404/403 bug where, after connecting via OAuth and then adding an API key, materialization used the wrong team's credential.

Supersedes #214 (closed) — rebuilt from scratch, with every OCS API call verified against the real `dimagi/open-chat-studio` source (no invented endpoints).

### What changed (model)
- **`TenantConnection`** is now a first-class credential: one row per credential a user adds (one OAuth login, or one API key). Replaces the per-chatbot `TenantCredential` (deleted).
- **`TenantMembership`** gains `connection` (FK), per-chatbot `team_slug`/`team_name`, and `archived_at` (soft-delete).
- Constraint: **one OAuth connection per `(user, provider)`**; unlimited API keys.

### Behavior
- **Resolution is fail-closed.** A chatbot resolves via its connection. For OAuth, if the live OIDC `team` claim (`SocialAccount.extra_data["team"]`) no longer matches the chatbot's `team_slug`, it returns `None` instead of silently using the wrong-team token — this is the actual fix for the reported bug and its re-login variant.
- **API-key team auto-detection** via `GET /api/sessions/?page_size=1` (the only API-key-reachable source of team name/slug), falling back to a user-typed team name when the team has no sessions.
- **Remove = archive.** Removing a connection (or disconnecting an OAuth provider) hides its chatbots and retains their conversations/materialized data; re-adding the credential restores them.
- Works for **OCS and CommCare** API keys; one OCS OAuth team in v1 (additional teams via API keys; the model already stores team identity so N-team OAuth is a later additive change).

### Endpoints
`/api/auth/connections/` (GET grouped list, POST add) and `/api/auth/connections/<id>/` (PATCH rotate, DELETE remove). The Connections page renders one card per connection with its team label and chatbots; OAuth connections are display-only (managed via the OAuth Providers section).

### Verification
- 1133 backend tests pass (8 skipped); ruff clean; migrations consistent; frontend `tsc`/eslint clean.
- New tests cover the end-to-end reported bug, archive/restore round-trip, multi-key isolation, OAuth disconnect archival, team auto-detect + fallback, and the data migration mapping.
- Design spec: `docs/superpowers/specs/2026-06-05-multi-team-ocs-connections-design.md`.


tests/test_ocs_connections.py
```diff
@@ -0,0 +1,527 @@
...
+    connection per credential with ciphertext preserved.
+
+    TenantCredential was dropped by 0008, so we recreate just that one historical
+    table (the other tables exist at head and the 0006-state models map to them),
+    seed it, run forward(), and drop it again.
+    """
+    import importlib
+
+    from django.db import connection
+    from django.db.migrations.loader import MigrationLoader
```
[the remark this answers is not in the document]
bderenzi: Fixed in d89a0b4 — all inline imports in this file are hoisted to the module-level block (module-level imports verified used; no try/except-ImportError or pre-django.setup() exceptions apply here).

bderenzi:
I started from scratch. I read through the spec and plan and tested this locally with success. Note that #221 still exists, but that's a separate PR in the future.

apps/users/models.py
```diff
@@ -177,22 +150,65 @@ class TenantCredential(models.Model):
...
+    )
+    connection = models.ForeignKey(
+        "TenantConnection",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="memberships",
+    )
+    team_slug = models.CharField(max_length=255, blank=True, default="")
+    team_name = models.CharField(max_length=255, blank=True, default="")
```
snopoke: These are very specific to one provider - do they need to be first class fields on the model or could we push these to a JSON field?

apps/users/services/credential_resolver.py
```diff
@@ -54,60 +54,75 @@ async def aget_social_token(user, provider: str) -> SocialToken | None:
...
+    SocialAccount's ``extra_data``. When they differ we must not use this token
+    (it has moved to another OCS team) — fail closed.
+    """
+    if not membership.team_slug:
+        return False
+    current = (getattr(token_obj.account, "extra_data", None) or {}).get("team")
+    return bool(current) and current != membership.team_slug
+
+
 def resolve_credential(membership) -> dict | None:
```
snopoke: I don't think this function is actually used anywhere except in tests. Prod code uses the async version.

apps/users/services/ocs_team.py
```diff
@@ -0,0 +1,58 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _base_url() -> str:
+    return getattr(settings, "OCS_URL", "https://www.openchatstudio.com").rstrip("/")
+
+
+async def _team_from_sessions(headers: dict, base_url: str) -> tuple[str, str] | None:
+    """GET /api/sessions/?page_size=1 -> (slug, name), or None when no sessions."""
```
snopoke, line 28: New API coming soon: https://github.com/dimagi/open-chat-studio/issues/3586

apps/users/views.py
```diff
@@ -294,8 +329,12 @@ async def tenant_credential_detail_view(request, membership_id):
             status=400,
         )
 
+    # Verify the new key still has access to one of this connection's chatbots.
+    sample = await conn.memberships.select_related("tenant").afirst()
+    if sample is None:
+        return JsonResponse({"error": "Connection has no chatbots to verify against"}, status=400)
```
snopoke: Wording here shouldn't be OCS specific

apps/users/models.py
bderenzi: Done in c98a5c5. `team_slug`/`team_name` now live in a `provider_metadata` JSONField on `TenantMembership` instead of as columns. I kept thin `team_slug`/`team_name` property accessors (get+set) so the keys stay in one place — the OAuth fail-closed check still reads `membership.team_slug`, and existing reads/writes plus `create(team_slug=...)` keep working unchanged. Verified nothing filters/orders on these columns, so dropping them is safe; the API response shape is identical.

apps/users/services/credential_resolver.py
bderenzi: Confirmed — sync `resolve_credential` was only referenced by two tests; prod uses `aresolve_credential`. Removed it in c98a5c5, along with the equally-dead sync `get_social_token` (`aget_social_token` is the live one), and converted the two tests to async against `aresolve_credential` so the api_key / no-connection coverage is retained.

apps/users/services/ocs_team.py
bderenzi: Noted. Added a TODO in `ocs_team.py` referencing issue 3586 (c98a5c5) so the sessions-based detection and its no-sessions fallback get swapped for the dedicated team endpoint once it ships.

apps/users/views.py
bderenzi: Fixed in c98a5c5 — the rotate-verify error now reads "Connection has no linked data sources to verify against" instead of "chatbots", since this endpoint also serves CommCare API keys.

PR #224, opened 2026-06-10
Survive dead worker DB connections and surface stuck materializations

bderenzi:
## Summary

Prod incident (2026-06-09/10): the worker's Django ORM connection died after the daily RDS backup window (an RDS minor-version upgrade restart preceded it by a day) and was never re-established. The worker has no HTTP request cycle, so Django's `close_old_connections()` hook never ran; the dead connection was reused forever and **every background task failed for ~22h** with `psycopg.OperationalError: the connection is closed` — including `expire_stale_thread_jobs`, the janitor that exists to rescue jobs stranded by exactly this. Users saw a "Preparing…" spinner that never progressed and never errored.

## Changes

**1. `ensure_fresh_db_connections` (config/procrastinate.py)** — every procrastinate task now runs `close_old_connections()` on the async-ORM executor thread before its body, so a connection that died between jobs is discarded and the next ORM call reconnects. `tests/test_worker_db_resilience.py` kills the underlying psycopg connection behind Django's back and proves a wrapped task recovers; a registry test enforces that every task in `apps.workspaces.tasks` carries the wrapper so new tasks can't reintroduce the bug. The pipeline progress updater thread (`asyncio.to_thread`, its own thread-local connection) gets the same treatment at entry.

**2. `reconcile_stale_thread_job` (apps/workspaces/tasks.py)** — the janitor's per-job logic, extracted and shared. New behavior: a stale PENDING ThreadJob whose procrastinate job **failed/aborted** flips straight to FAILED with a user-facing `error_summary` and synthetic chat message, instead of deferring an agent resume that has no result to narrate.

**3. API-side backstop (apps/workspaces/api/jobs_views.py)** — `active_jobs_view` (polled every few seconds by the frontend) reconciles stale active jobs itself. The API process stayed healthy throughout the incident; the watchdog no longer lives only in the failure domain it is watching. Fresh jobs are untouched (no extra queries on the hot path); the frontend's existing failure card (`recent_terminations` → `error_summary` + retry) renders within one poll of the 10-minute staleness threshold. No frontend changes needed.

## Deliberately not included

- **Task-level retry on transient DB errors**: `materialize_workspace`'s `finally` block defers the resume task on every attempt, so procrastinate retries would double-fire resumes and race the failure narration. Worth revisiting only with a redesign of the resume-defer placement.
- The June incident also surfaced that worker-side Sentry errors fired for ~22h unnoticed — alerting routing is an ops follow-up, not a code change here.

## Test plan

- `uv run pytest` — 1120 passed, 8 skipped (7 new tests: 3 connection-resilience, 1 janitor failed-flip, 3 API backstop)
- `uv run ruff check .` / `ruff format` clean on touched files


tests/test_jobs_endpoints.py
```diff
@@ -605,3 +605,134 @@ async def test_active_jobs_percent_null_when_rows_total_missing():
...
+# The janitor task runs in the worker process; when the worker itself is sick
+# (June 2026 incident: its DB connection died and every task — including the
+# janitor — failed for ~22h) nothing flips stuck ThreadJobs and the frontend
+# spins on "Preparing…" forever. The polling endpoint runs in the healthy API
+# process, so it reconciles stale active jobs itself before responding.
+# ---------------------------------------------------------------------------
+
+
+async def _make_stale_pending_job(email: str, job_id: int) -> tuple:
+    user = await sync_to_async(User.objects.create_user)(email=email, password="x")
```
snopoke: can we get rid of these `sync_to_async` calls?

bderenzi: Done in 3350415 — and took it further: that idiom predated the async ORM and had been copy-pasted across 12 test files (~380 call sites), so I converted the whole suite to native `acreate`/`alogin`/`arefresh_from_db`. Root cause of the idiom: our custom `UserManager` subclasses `BaseUserManager`, which doesn't have `acreate_user` (Django's lives on `UserManager`) — added a native `acreate_user` to the manager. The only remaining `sync_to_async` in tests is the worker-resilience test's DB-connection manipulation, which must run on the thread that owns the connection. A sweep confirmed all 8 production usages are legitimate bridges (external sync APIs, `transaction.atomic`, `close_old_connections` thread affinity).

apps/users/models.py
```diff
@@ -23,6 +23,19 @@ def create_user(self, email=None, password=None, **extra_fields):
         user.save(using=self._db)
         return user
 
+    async def acreate_user(self, email=None, password=None, **extra_fields):
```
snopoke, line 26: Django already has this method: https://docs.djangoproject.com/en/5.2/ref/contrib/auth/#django.contrib.auth.models.UserManager.acreate_user

config/procrastinate.py
```diff
@@ -5,6 +5,40 @@
...
+
+from asgiref.sync import sync_to_async
+from django.db import close_old_connections
 from procrastinate.contrib.django import app
 
-__all__ = ["app"]
+__all__ = ["app", "ensure_fresh_db_connections"]
+
+
+def ensure_fresh_db_connections(func):
```
snopoke: What about creating a custom task decorator instead of having 2 decorators which need to be applied to every task?

See https://procrastinate.readthedocs.io/en/stable/howto/advanced/middleware.html

bderenzi: Done in b7b61af — exactly that pattern (the docs' "middleware" is this custom-decorator approach; procrastinate has no built-in hook as of 3.8.1). One `@task` decorator now wraps the body and delegates registration to `app.task`, and the registry test enforces that everything in `apps.workspaces.tasks` goes through it.

apps/users/models.py
bderenzi: Close but not quite usable here: Django's `acreate_user` lives on `UserManager`, and our manager subclasses `BaseUserManager` because the model uses email as USERNAME_FIELD — `UserManager`'s API is username-first (`create_user(username, email=None, ...)`) and its `_create_user` normalizes/requires username, so inheriting it would mean overriding both the sync and async pair anyway, plus we'd lose the store-NULL-email behavior. The custom-manager-on-BaseUserManager shape is the documented Django pattern for email-login models, and Django itself implements `acreate_user` as a duplicated async twin of `_create_user` (django/contrib/auth/models.py:166) rather than a wrapper — this mirrors that. Happy to switch if you see a cleaner subclassing route.

snopoke (approved):

bderenzi:
## Upstream context: this is a known procrastinate gap, with a fix in flight

The dead-connection failure this PR works around is procrastinate-org/procrastinate#1134 (open since July 2024) — identical traceback (`psycopg.OperationalError: the connection is closed` on a task's first ORM call), and the thread converged on exactly the remedy used here: `close_old_connections()` before each task. One report from 2025-11 is nearly our incident verbatim (AWS RDS Proxy killing idle connections). The thread also confirms why `CONN_HEALTH_CHECKS=True` doesn't help: that machinery only runs in Django's request cycle, which a worker never enters.

Better still: procrastinate-org/procrastinate#1555 (maintainer-authored, 2026-06-02) adds this cleanup **natively** to the Django contrib — `close_old_connections()` before and after every task, covering both sync and async tasks. It's currently a draft (the author is reworking the registration-hook mechanism), so our decorator is the right fix today.

**Removal plan:** #225 tracks stripping the custom `task` decorator once the upstream fix is merged, released, and we've upgraded past it. Until then, overlap is harmless (`close_old_connections()` is idempotent). The decorator's docstring carries the same pointer so whoever touches it knows it's temporary.

config/procrastinate.py
```diff
@@ -5,6 +5,52 @@
...
+    TEMPORARY: this is a workaround for procrastinate-org/procrastinate#1134;
+    upstream PR #1555 adds the same cleanup natively in the Django contrib.
+    Once that merges and we upgrade past the release containing it, strip this
+    wrapper — see https://github.com/dimagi-rad/scout/issues/225.
+    """
+
+    def wrap(func):
+        @functools.wraps(func)
+        async def wrapper(*args, **kwargs):
+            await sync_to_async(close_old_connections)()
```
snopoke: 2 things I notice from https://github.com/procrastinate-org/procrastinate/pull/1555/:

1. use `thread_sensitive=True` to make sure the function runs in the same thread context
2. call `django.db.reset_queries` after the task function (mostly for when `DEBUG=True`)

bderenzi: Both done in fe93603. `thread_sensitive=True` is the asgiref default so we were already getting it, but it's load-bearing here (cleanup must run on the thread that owns the connections) so it's now explicit with a comment. And the decorator now cleans up **after** the body too — `close_old_connections()` so a worker thread doesn't hold a connection between jobs (new test pins this), plus `reset_queries()` for the DEBUG-only query log. The wrapper now mirrors #1555's before/after shape exactly, which should also make the eventual swap-out (#225) a behavioral no-op.

snopoke (approved):

PR #229, opened 2026-06-10
fix(workspaces): surface view-schema build failures to agent, MCP and status API

bderenzi:
## Problem (prod incident 2026-06-10)

When the post-materialization `build_view_schema` fails for a multi-tenant workspace, the failure is swallowed: the resume task tells the agent "Materialization just completed (status=completed). Per-tenant: …" while `get_schema_status` reports a bare `not_provisioned` and the workspace status API shows nothing. Faced with "completed" + "not_provisioned", the agent re-ran the entire materialization (which can never fix a view-build failure) and finally told the user to contact a system administrator. The real error was never visible anywhere.

## Fix

- `WorkspaceViewSchema.last_error` (new TextField + migration): `build_view_schema` stores the failure message (truncated to 500 chars) on the FAILED path and clears it on success.
- `materialize_workspace` returns a `view_schema: {ok, error}` outcome alongside the per-tenant results.
- `resume_thread_after_materialization`: when runs completed/partial but the multi-tenant view schema is not ACTIVE, the system-resume message now states that per-tenant data loaded **but the query layer failed to build**, quotes the error, and instructs the agent **not** to re-run materialization. The ThreadJob flips to FAILED (with a user-facing `error_summary`) instead of COMPLETED, so the UI shows an honest failure instead of a spinner-then-confusion.
- `get_schema_status` (MCP): a FAILED view schema now returns `state: "failed"` + `error` instead of `not_provisioned` (additive response shape).
- `_derive_schema_status` (status API): FAILED view schema → `"failed"` (frontend consumes this in exactly one place, `workspaceDataState()`, with equality checks and a safe fallback — the fallback would have mislabeled a failed workspace as "ready", so the frontend maps `"failed" → "empty"`).

## Tests

Resume-prompt content for the failed case (+ regression for the ACTIVE case), MCP failed-state response, status-API derivation, `last_error` store/clear, migration applies.

`uv run pytest` full suite green (1184 passed) on the integration branch combining the six incident-fix PRs. Test-file overlap with the truncation PR is union-trivial; I'll rebase whichever lands second.


snopoke (approved):
Merge conflicts need to be resolved.

Issue #238, opened 2026-06-15
[arch] Recipe runner signature drift -- feature 100% dead since March


bderenzi:
RecipeRunner calls build_agent_graph with a removed kwarg -> TypeError -> 500 -> RecipeRun stranded RUNNING forever. Triple drift (kwarg + initial-state + result-extraction). Restores the feature to working-as-designed; the redesign question is separate (see content-satellite-redesign).

## Findings (1)

- [ ] **`00#1`** BROKEN-NOW/correctness (r11, accidental) -- Recipe execution 100% broken: RecipeRunner calls build_agent_graph with removed tenant_membership signature

<details><summary>Evidence (claim / chain / files)</summary>

### `00#1` Recipe execution 100% broken: RecipeRunner calls build_agent_graph with removed tenant_membership signature
**Claim:** RecipeRunner._build_graph calls build_agent_graph(tenant_membership=..., user=..., checkpointer=None) but the signature is build_agent_graph(workspace, user, checkpointer, mcp_tools, oauth_tokens) with no **kwargs and no tenant_membership param (drift introduced by e26cd75, 2026-03-10) — TypeError on every recipe run. The graph build happens outside the try block, so the API returns HTTP 500 and the RecipeRun row created beforehand is stranded RUNNING forever (no janitor reconciles RecipeRun). Triple drift even if the kwarg were fixed: initial_state uses pre-rename keys (tenant_id/tenant_name/tenant_membership_id, no workspace_id/thread_id) absent from AgentState, so MCP injection passes workspace_id='' and every data tool fails with VALIDATION_ERROR (3 in a row trips the escalation breaker); mcp_tools is never loaded/passed; and execute() calls sync graph.invoke() on an async-only graph (execute_async has zero production callers). All recipe tests @patch build_agent_graph with Mock/AsyncMock, masking the 3-month-old break; test_recipe_runner.py is four @skip placeholders.

[...]

Issue #234, opened 2026-06-15
[arch] Real (unmocked) chat<->MCP contract test + frontend test infra


bderenzi:
Stand up FastMCP in-process with a real client, no mocks, on the highest-churn seam. Would have caught the recipe breakage, toolCallId mismatch, get_metadata card bug, and onboarding 404 as a CLASS. Includes frontend unit-test infra and retiring the mocks that pin dead seams.

## Findings (5)

- [ ] **`10#4`** DEBT/correctness (r2, accidental) -- Chat-to-MCP wire has zero unmocked coverage; tests mock the exact seams that hid production bugs
- [ ] **`10#5`** DEBT/velocity (r1, accidental) -- Frontend has no unit-test infrastructure; e2e runs nowhere; post-deploy smoke tests exist but are wired to nothing
- [ ] **`12#0`** DEBT/velocity (r1, accidental) -- Test suite is structurally blind to the known incident classes: mocks pin the exact dead seams (FutureApp current_app, oauth_tokens, FAILED-vs-CANCELLED, stale-cursor selection, RUNNING false-failure), affirms the pg_catalog disclosure as contract, and tests no role enforcement
- [ ] **`02#6`** DEBT/correctness (r3, accidental) -- Prompt instructs run_materialization with a pipeline= parameter the tool no longer accepts (single-tenant branch); recurring prompt-vs-contract drift
- [ ] **`07#0`** DEBT/velocity (r1, accidental) -- Context-injection contract relies on both LangChain and FastMCP skipping arg validation

<details><summary>Evidence (claim / chain / files)</summary>

### `10#4` Chat-to-MCP wire has zero unmocked coverage; tests mock the exact seams that hid production bugs
[...]

PR #274, opened 2026-06-15
test: real unmocked chat<->MCP contract test (arch #234)

bderenzi:
## What this builds

A **real, unmocked** contract test for the chat↔MCP wire — the highest-churn seam that, per arch finding **10#4**, had zero unmocked coverage because every existing test mocks exactly the seam that hid production bugs (`MultiServerMCPClient` patched, `get_mcp_tools` faked, `MCP_TOOL_NAMES` asserted against itself).

The test stands up the **real** `mcp_server.server.mcp` FastMCP instance over the official MCP SDK's in-memory transport (`mcp.shared.memory.create_connected_server_and_client_session`) and loads its tools with the **same** `langchain_mcp_adapters.load_mcp_tools` loader the agent graph uses (`apps/agents/mcp_client.get_mcp_tools`). Nothing on the wire is mocked: the LangChain tool, the MCP protocol round-trip, FastMCP arg parsing, and the tool handler all run for real. One test even drives the **production injecting tool node** (`_make_injecting_tool_node`) inside a compiled `StateGraph`, exercising the full `chat → graph → mcp_client → MCP-tool` path end to end.

Files:
- `tests/test_chat_mcp_contract.py` — the contract test
- `docs/superpowers/plans/2026-06-15-chat-mcp-contract-test.md` — plan

## Contracts asserted (passing)

- **Tool-set drift detector** — tools advertised over the real wire == `MCP_TOOL_NAMES` ∪ context-free tools. Catches the rename/add/remove drift that the self-referential `MCP_TOOL_NAMES` test cannot (the recipe-class bug).
- **Injected-param presence** — every tool in `MCP_TOOL_NAMES` advertises `workspace_id` on its real `inputSchema`; `run_materialization` advertises all four injected params (`workspace_id/user_id/thread_id/tool_call_id`) and **no** `pipeline`.
- **Real round-trip shapes** — `list_pipelines` and `get_schema_status` round-trip and return the documented envelope (`success` + documented `data.*`) the agent/frontend depend on.
- **Server-side trust boundary** — empty `workspace_id` is rejected server-side (`VALIDATION_ERROR`); `run_materialization` enforces membership/thread ownership server-side rather than trusting the LLM.
- **End-to-end injection** — `workspace_id` flows from agent state through the real injecting tool node into the real MCP server (no mock between state and server).

## xfailed contracts (correct assertion, tracked)

- **Prompt-vs-tool-schema drift** (`@pytest.mark.xfail(strict=False)`) — the single-tenant `## Data Availability` prompt still instructs `run_materialization(pipeline="…")`, but the tool advertises no `pipeline` param. Asserts the **correct** contract (prompt must not reference a param the tool lacks). **Blocked by #238**; will xpass when fixed. (finding 02#6)

This guardrail addresses the finding class as a whole: prompt-vs-tool-schema drift (02#6), arg-validation-by-omission (07#0), and the unobservable wire (10#4) — the same class that produced the recipe breakage, the toolCallId mismatch, the get_metadata "0 tables" card, and the onboarding 404.

## Deferred

Frontend unit-test infrastructure (finding **10#5**) is **out of scope** here and deferred to a follow-up, as noted in the plan and on #234.

## Local verification

Full backend suite green against a throwaway managed DB: **1194 passed, 8 skipped, 22 deselected, 1 xfailed, 0 failures**. Throwaway DB and all leftover `ws_*_ro` roles dropped after the run (residual `pg_roles` diff empty).

Partially addresses #234.

### DIFF
--- a/docs/superpowers/plans/2026-06-15-chat-mcp-contract-test.md
+++ b/docs/superpowers/plans/2026-06-15-chat-mcp-contract-test.md
@@ -1,0 +1,72 @@
+# Plan: Real (unmocked) chat<->MCP contract test (arch #234, partial)
+
+## Problem (from issue #234)
+
+The chat->MCP wire has ZERO unmocked test coverage. Every existing test mocks
+exactly the seam that hid production bugs:
+
+- `test_mcp_client` patches `MultiServerMCPClient`
+- `test_mcp_chat_integration` mocks `get_mcp_tools` with hand-written fakes whose
+  schemas can drift from `server.py`
+- `test_mcp_server` / `test_mcp_tenant_tools` call handlers in-process with the DB
+  mocked at the ORM/psycopg boundary
+- `test_agent_graph` asserts `MCP_TOOL_NAMES` against itself
+
+This masking pattern hid: the recipe signature drift, the toolCallId mismatch, the
+`get_metadata` "0 tables" card, and the onboarding 404 — as a CLASS. Findings:
+10#4, 12#0, 02#6, 07#0.
+
+## Approach: in-process FastMCP over a real MCP wire
+
+The official `mcp` SDK ships an in-memory transport,
+`mcp.shared.memory.create_connected_server_and_client_session(server)`, which runs
+the REAL `mcp_server.server.mcp` FastMCP instance in a task group and hands back a
+real `ClientSession` connected over anyio memory streams. `langchain_mcp_adapters.tools.load_mcp_tools(session)` then loads the server's tools as the same LangChain
+tools the agent graph uses — over the real protocol, no mocks on the wire.
+
+Verified by spike: 11 tools load; `list_pipelines` round-trips and returns the real
+JSON envelope; `session.list_tools()` returns the real `inputSchema` per tool.
+
+## Contracts to assert (highest value subset; depth over breadth)
+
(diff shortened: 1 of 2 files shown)


bderenzi:
Addressed the auto-review CLAUDE.md violation in `tests/test_chat_mcp_contract.py` in 919ed23.

Replaced every `sync_to_async(...)`-wrapped bare ORM call with native async ORM (none of the flagged sites was a `transaction.atomic()` block, so the documented exception did not apply):

- `sync_to_async(Tenant.objects.create)(...)` → `await Tenant.objects.acreate(...)`
- `sync_to_async(Workspace.objects.create)(...)` → `await Workspace.objects.acreate(...)`
- `sync_to_async(WorkspaceMembership.objects.create)(...)` → `await WorkspaceMembership.objects.acreate(...)`
- `_make_user` is now `async def` calling `await get_user_model().objects.acreate_user(...)`, awaited at all call sites
- dropped the now-unused `sync_to_async` import

Test contracts/assertions and the in-process FastMCP setup are unchanged.

Verified locally (Postgres.app, `config.settings.test`): `pytest tests/test_chat_mcp_contract.py -v` → **8 passed, 1 xfailed, 0 failures**; `ruff check` clean.

PR #277, opened 2026-06-15
Fix recipe runs crashing with SynchronousOnlyOperation

### LINKED ISSUE #276: SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
Sentry Issue: [SCOUT-DJANGO-1P](https://dimagi.sentry.io/issues/7553662856/?referrer=github_integration)

```
KeyError: 'workspace'
  File "django/db/models/fields/related_descriptors.py", line 243, in __get__
    rel_obj = self.field.get_cached_value(instance)
  File "django/db/models/fields/mixins.py", line 37, in get_cached_value
    return instance._state.fields_cache[self.cache_name]

SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
(8 additional frame(s) were not displayed)
...
  File "apps/recipes/api/views.py", line 108, in post
    run = runner.execute()
  File "apps/recipes/services/runner.py", line 191, in execute
    graph = async_to_sync(self._build_graph)()
  File "apps/recipes/services/runner.py", line 107, in _build_graph
    .filter(workspace=self.recipe.workspace)

Error running recipe 3065d97a-47be-4aaf-be65-4c6598530c2e
```

jjackson:
Fixes #276 (Sentry SCOUT-DJANGO-1P).

## Problem
Running a recipe crashed with `SynchronousOnlyOperation: You cannot call this from an async context`.

`RecipeRunView.post` (a sync DRF view) loads the recipe with `Recipe.objects.get(pk=..., workspace=...)`. Filtering *by* the workspace FK does **not** cache the related object on the instance. `execute()` then runs `_build_graph` via `async_to_sync` — inside an async event loop — where `.filter(workspace=self.recipe.workspace)` evaluated the uncached forward FK (`KeyError: 'workspace'` cache miss), triggering a synchronous lazy DB load. Django forbids sync ORM in an async context, hence the crash.

## Solution
Reference the FK by its already-loaded column attribute `workspace_id` instead of the lazy-loading `workspace` object — equivalent lookup, zero DB I/O, async-safe. Applied to `_build_graph` (the reported crash) and to `execute_async`, which carried the identical latent access.

## Why we can be confident
Added a regression test that re-fetches the recipe (uncached FK, mirroring the view) and drives `_build_graph` through `async_to_sync`. It reproduces the exact `SynchronousOnlyOperation` at `runner.py:107` before the fix and passes after. Full recipe suite (49 passed) and `ruff` clean.

### DIFF
--- a/apps/recipes/services/runner.py
+++ b/apps/recipes/services/runner.py
@@ -102,9 +102,13 @@ class RecipeRunner:
             return self._provided_graph
 
         if self._graph is None:
+            # Reference the FK by its already-loaded column (workspace_id) rather
+            # than self.recipe.workspace: this coroutine runs inside async_to_sync,
+            # and a lazy FK load would issue a sync DB query, raising
+            # SynchronousOnlyOperation in the async context.
             workspace_tenant = (
                 await WorkspaceTenant.objects.select_related("tenant")
-                .filter(workspace=self.recipe.workspace)
+                .filter(workspace_id=self.recipe.workspace_id)
                 .afirst()
             )
             tenant = workspace_tenant.tenant if workspace_tenant else None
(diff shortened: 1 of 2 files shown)


bderenzi:
Thanks for flagging, Jon and Jon's Claude. This is a symptom of #238 . I'm not going to approve or merge this PR because (1) the fix shouldn't keep the synchronous path (see the async-first convention) and (2) there are a number of other issues with recipes that aren't covered by this PR. I'm working to get all the issues addressed over the next couple of days.

bderenzi:
Closing as stale/superseded (per discussion above). The SynchronousOnlyOperation crash this targets (#276) already landed on main via #278; the remaining recipe-materialization fix is tracked in #294. Not merging this one.

PR #294, opened 2026-06-17
fix(recipes): materialize headlessly instead of crashing on the synthetic thread_id

bderenzi:
## Summary

Recipe runs that needed fresh data were crashing in production (SCOUT-DJANGO-1R / 1S). The recipe agent followed its prompt and called `run_materialization`, which cast the recipe's **synthetic** `thread_id` (`recipe-run-<uuid>`) to a `Thread` UUID and raised `ValidationError: "…" is not a valid UUID`.

The real problem is a **category error**: `run_materialization` is an *interactive, fire-and-resume* tool that only works for a caller owning a persisted chat `Thread` + checkpointer + async-resume loop. A recipe is a *one-shot, headless, thread-less, checkpointer-less* run. This PR makes recipe execution explicitly **headless**: it gets a *blocking* materialize that runs the pipeline inline and returns, then continues building the dashboard — no Thread, no checkpointer, no resume needed (a headless run has no interactive turn to protect).

**Verified against live prod:** the failing id `f3be369b-…` is a `RecipeRun` (status=`failed`), there is **no** `chat_thread` row with that id, and the stored step error is the exact crash — so patching the UUID alone was never enough (the Thread lookup would still miss).

## Root cause

| When | Commit | Effect |
|---|---|---|
| Feb 5 | `befce76` | Recipe runner born with `checkpointer=None` + synthetic `thread_id`. Harmless — no thread-coupled tools yet. |
| May 20–21 | `af2e353`, `aa908d1` | `run_materialization` becomes fire-and-ack + adds `Thread.objects.filter(id=thread_id)`. **Arms the bomb.** |
| Jun 16 | `6e866de` (arch #238) | Fixes the recipe runner so recipes actually run for the first time → first materializing recipe **detonates** the latent crash. |

Systemic root: the agent graph had an **implicit, unenforced contract** — every caller must invoke it on behalf of a real persisted `Thread` + checkpointer. Chat honored it; the recipe runner (the only headless, only contract-violating front-end) did not. The contract lived only in chat's helpers, invisible at the call boundary.

## The fix (defense in depth + systemic close)

1. **Guard** (`mcp_server/server.py`): `run_materialization` rejects a non-UUID `thread_id` with a clean validation error instead of a 500 — stops the crash for *any* caller.
2. **Headless blocking tool** (`apps/agents/tools/materialization_tool.py`): runs `materialize_workspace_core` inline and blocks until done (core extracted from the `materialize_workspace` task, which keeps its resume deferral).
3. **Explicit mode** (`apps/agents/graph/base.py`): `build_agent_graph(interactive=...)`. Headless swaps the MCP fire-and-ack `run_materialization` for the blocking tool and emits "this blocks and returns" prompt guidance instead of "end your turn, you'll be resumed". **Closes the bug class** — any future headless front-end is safe by construction.
4. **Headless execution** (`apps/recipes/`): the runner builds the headless graph; recipe execution moves to a background `run_recipe` task (it could block on materialization, which must not hold the HTTP request open). The endpoint returns **202** with a PENDING run; the frontend polls to completion.

## Test plan

- Backend: **1303 passed, 5 skipped, 0 failures** (full suite).
- Frontend: **eslint clean**, **`vite build` exit 0**.
- Task-0 regression test red-green verified (reproduces the exact prod `ValidationError`, then passes).
- Capstone e2e (`tests/test_recipe_materialization_integration.py`) drives the **real** graph wiring: a recipe agent calling `run_materialization` completes, the blocking core runs with the run's `job_id`, and **no `Thread`/`ThreadJob`** is created.
- Live prod DB confirmation (above).

## Deploy notes

- `run_recipe` runs on a dedicated `recipes` queue. The default worker processes all queues, so it works as-is; if recipe + chat-materialization load contend, run a dedicated worker (`--queues recipes`) or raise worker concurrency.
- API contract change: `POST .../recipes/<id>/run/` now returns **202 + PENDING run** (was 201 + completed). The frontend already polls.

### DIFF
--- a/apps/agents/graph/base.py
+++ b/apps/agents/graph/base.py
@@ -36,6 +36,7 @@ from apps.agents.prompts.artifact_prompt import ARTIFACT_PROMPT_ADDITION
 from apps.agents.prompts.base_system import BASE_SYSTEM_PROMPT
 from apps.agents.tools.artifact_tool import create_artifact_tools
 from apps.agents.tools.learning_tool import create_save_learning_tool
+from apps.agents.tools.materialization_tool import create_materialization_tool
 from apps.agents.tools.recipe_tool import create_recipe_tool
 from apps.knowledge.services.retriever import KnowledgeRetriever
 from apps.workspaces.models import (
@@ -143,18 +144,21 @@ _system_prompt_cache: dict[str, tuple[str, float]] = {}
 _SYSTEM_PROMPT_TTL = 60  # 60 seconds — short to limit staleness from knowledge/schema changes
 
 
-def _system_prompt_cache_key(workspace, user) -> str:
+def _system_prompt_cache_key(workspace, user, interactive: bool = True) -> str:
     """Build a cache key from workspace + user properties that affect the prompt.
 
     Includes user.id because _fetch_schema_context scopes TenantMetadata
     lookup to the specific user. Includes workspace.system_prompt hash
-    so edits invalidate immediately.
+    so edits invalidate immediately. Includes ``interactive`` because the
+    materialization guidance differs between interactive (fire-and-resume) and
+    headless (blocking) runs — caching one for the other would mislead the agent.
     """
     prompt_hash = hashlib.md5(
         (workspace.system_prompt or "").encode(), usedforsecurity=False
     ).hexdigest()[:8]
     user_id = getattr(user, "id", "anon")
-    return f"{workspace.id}:{user_id}:{prompt_hash}"
+    mode = "i" if interactive else "h"
+    return f"{workspace.id}:{user_id}:{prompt_hash}:{mode}"
 
 
 def _render_compact_schema(tables: list[dict], last_materialized_at: str | None) -> str:
@@ -217,11 +221,27 @@ def _render_full_schema(
     return "\n".join(lines)
 
 
-async def _fetch_schema_context(tenant, user) -> str:
+# Guidance injected for HEADLESS (non-interactive) runs — e.g. recipe execution.
+# Such runs have no chat Thread/checkpointer and no async-resume path, so the
+# agent must NOT "end its turn and wait": the headless run_materialization tool
+# BLOCKS and returns when loading finishes, and the agent continues in the same
+# run. (Deliberately avoids the substring "end your turn".)
+_HEADLESS_MATERIALIZE_GUIDANCE = (
+    "No data has been loaded yet. Call `run_materialization` to load it. This "
+    "tool BLOCKS and returns a status summary once loading finishes — keep "
+    "working in the same run rather than stopping. After it returns "
+    "`status: completed`, continue with the requested analysis; the data is ready."
+)
+
+
+async def _fetch_schema_context(tenant, user, interactive: bool = True) -> str:
     """Fetch database schema state and build a ## Data Availability prompt section.
 
     Tries to build a full schema block (tables + columns). Falls back to a compact
     block (tables + row counts only) if the full text exceeds SCHEMA_CONTEXT_CHAR_BUDGET.
+
+    ``interactive`` selects the materialization guidance: fire-and-resume (chat)
+    vs blocking (headless recipe runs).
     """
     ts = await TenantSchema.objects.filter(
         tenant=tenant,
@@ -232,6 +252,8 @@ async def _fetch_schema_context(tenant, user) -> str:
     pipeline_config = registry.get_by_provider(tenant.provider)
 
     if ts is None:
+        if not interactive:
+            return _HEADLESS_MATERIALIZE_GUIDANCE
         # NB: no `pipeline=` argument — run_materialization takes no pipeline
         # parameter (routing moved into materialize_workspace per-provider) and
         # all of its real params are injected server-side and hidden, so the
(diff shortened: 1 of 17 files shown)


apps/agents/tools/materialization_tool.py
```diff
@@ -0,0 +1,90 @@
...
+
+        Blocks until loading completes, then returns a status summary. Call this
+        before querying when no data has been loaded yet. After it returns
+        ``status: completed``, continue with the requested analysis in the same
+        run — the data is ready.
+        """
+        # Imported lazily: apps.workspaces.tasks imports apps.agents.graph.base
+        # (build_agent_graph) at module level, so a module-level import here
+        # would create an agents<->workspaces.tasks import cycle.
+        from apps.workspaces.tasks import materialize_workspace_core
```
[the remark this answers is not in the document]
bderenzi: This one is a genuine exception, not an oversight. Verified empirically — moving it to module level fails at startup:

```
ImportError: cannot import name 'materialize_workspace_core' from partially
initialized module 'apps.workspaces.tasks' (most likely due to a circular import)
```

in every import order. The cycle is `apps.agents.graph.base → apps.agents.tools.materialization_tool → apps.workspaces.tasks → apps.agents.graph.base` (`tasks.py` imports `build_agent_graph` for the resume path; `base.py` imports this tool factory for `_build_tools`).

This mirrors the established pattern in `apps/agents/tools/recipe_tool.py:98`, which imports `apps.recipes.models` inside the tool body for the same reason. Breaking the cycle structurally would mean extracting the materialization core + its task-coupled helpers into a new base-independent module — a sizeable, riskier refactor out of scope for this fix.

I've expanded the comment to record the verified cause and the precedent (commit pushed). Keeping the lazy import.

apps/agents/graph/base.py
```diff
@@ -246,6 +268,8 @@ async def _fetch_schema_context(tenant, user) -> str:
         )
 
     if ts.state == SchemaState.MATERIALIZING:
+        if not interactive:
+            return _HEADLESS_MATERIALIZE_GUIDANCE
```
snopoke: Won't this result in a 2nd materialization being kicked off (since we're telling the agent that no data has been loaded yet). In the interactive path we explicitly tell it that one is in progress and it should not start another one.

apps/agents/graph/base.py
```diff
@@ -341,6 +365,8 @@ async def _fetch_multi_tenant_schema_context(workspace, user) -> str:
         ).afirst()
 
     if active_run is not None or (vs is not None and vs.state == SchemaState.MATERIALIZING):
+        if not interactive:
+            return _HEADLESS_MATERIALIZE_GUIDANCE
```
snopoke: similar comment to above about in progress materialization prompt

apps/recipes/services/runner.py
```diff
@@ -154,14 +169,21 @@ async def execute_async(self) -> RecipeRun:
...
 
-        self._run = await RecipeRun.objects.acreate(
-            recipe=self.recipe,
-            status=RecipeRunStatus.RUNNING,
-            variable_values=self.variable_values,
-            step_results=[],
-            started_at=timezone.now(),
-            run_by=self.user,
-        )
+        if self._run is None:
```
snopoke: when would `self._run` be `None`?

apps/agents/graph/base.py
bderenzi: Good catch — that was a real bug. The MATERIALIZING branch was returning the same `_HEADLESS_MATERIALIZE_GUIDANCE` ('no data loaded, call run_materialization') as the no-data branch, so the headless agent would start a 2nd parallel materialization.

Fixed two ways (1f27f9c):
1. New `_HEADLESS_MATERIALIZE_IN_PROGRESS_GUIDANCE` for the MATERIALIZING branches — still routes to `run_materialization` (it's the only headless path to data) but no longer says 'no data loaded'.
2. The headless tool now calls a new `materialize_workspace_blocking()` that **waits for any in-progress MaterializationRun on the workspace's tenants to clear** before running its own — so even if the agent does call it, no parallel run against the same tenant schema happens. (A headless run has no async-resume, so it can't just 'wait and be resumed' like the interactive path — the tool waits inline instead.)

apps/agents/graph/base.py
bderenzi: Same fix as the single-tenant branch above (1f27f9c): this multi-tenant MATERIALIZING branch now returns `_HEADLESS_MATERIALIZE_IN_PROGRESS_GUIDANCE`, and the headless `run_materialization` tool waits out any in-progress run rather than starting a parallel one.

apps/recipes/services/runner.py
bderenzi: You're right — after the background-task migration, production always sets it: the view creates the `RecipeRun(PENDING)` and `run_recipe` passes it in, so `self._run` was only ever None in direct/test callers. Made `run` a required keyword arg and dropped the create-in-`execute_async` branch (350be58). Variable validation moved to a `RecipeRunner.validate_and_default` staticmethod so the view can validate + default before creating the run without constructing a runner; the TestRecipeRunner tests now create a PENDING run and pass it.

snopoke (approved):

PR #302, opened 2026-06-18
Connect form-aware staging: deliver-app schema → typed stg_visits columns + auto column notes

jjackson:
## What this does

Makes Scout **form-aware** for CommCare Connect data. Previously the agent saw `raw_visits.form_json` as an opaque JSON blob; now it sees typed, human-labeled staging columns with auto-generated notes — which lifts answer quality on **every** Connect opportunity.

Two pieces:

**1. Fetch the deliver-app form schema (M1).** `ConnectMetadataLoader` now calls Connect's `GET /export/opportunity/{id}/app_structure/?app_type=both` (Connect proxies to CommCare HQ server-side — no HQ credentials needed) and reuses the existing `_extract_form_definitions` to parse the `{learn_app, deliver_app}` HQ app JSON into `form_definitions`, stored in `TenantMetadata`. Degrades gracefully if the endpoint is unavailable.

**2. Stage `form_json` into typed/labeled columns (M2).**
- `connect_staging.py` generates a `stg_visits` dbt asset flattening `raw_visits.form_json` into typed columns (one per non-repeat question, cast by question type), plus `stg_visits__repeat_<group>` child assets for repeat groups — mirroring the existing `commcare_staging` conventions, with a shared `visit_column_map` as the single source of truth for column naming.
- `column_note_generator.py` auto-populates `TableKnowledge.column_notes` from the form questions (labels, types, choice lists). It's **merge-safe**: re-syncs preserve human-edited column notes and never clobber a human-written table `description` (Django 5 `create_defaults`). These notes already flow into the agent prompt via the existing `KnowledgeRetriever` — no retriever change needed.
- Wired into the materializer's transform phase as an additive `commcare_connect` branch (existing provider behavior untouched).

## Why land it on its own

This is independently valuable regardless of any downstream work: the agent now understands `muac` is a `Decimal` labeled "MUAC (cm)" instead of guessing at a JSON blob. It is the first step of a larger semantic-layer exploration (design + roadmap under `docs/superpowers/`), but it stands alone and improves Connect data handling today.

## Testing

~45 new tests, all green: app_structure parsing, discover persistence, staging SQL (typed columns + repeat groups + SQL-escaping), column-note merge/idempotency, and the materializer branch wiring (drives `run_pipeline` for `commcare_connect`). No model/migration changes.

Reviewed task-by-task plus a final whole-branch review (async correctness, SQL safety, orphan-sweep scoping, and staging↔notes column-name alignment incl. the collision case all verified).

## Deferred follow-up

- Repeat-group child tables (`stg_visits__repeat_*`) are staged but not yet annotated with column notes — a small follow-up, intentionally out of scope here.

### DIFF
--- a/docs/superpowers/specs/2026-06-18-cube-semantic-layer-auto-modeling-design.md
+++ b/docs/superpowers/specs/2026-06-18-cube-semantic-layer-auto-modeling-design.md
@@ -43,25 +43,35 @@ evaluation harness.
 ## 2. Goals / non-goals
 
 **Goals**
-- Get synthetic Connect data into Scout as a real data source (test fuel).
+- **Run the POC first and foremost against real CommCare Connect data.** Scout already has a
+  working Connect pipeline (`pipelines/connect_sync.yml`, `ConnectMetadataLoader`), so the POC
+  is not blocked on any external work.
 - **Auto-model** a Cube semantic layer from Connect/CommCare **app structure** (form schemas)
   + Scout's existing knowledge — not hand-written YAML.
 - Let the agent query governed measures, falling back to raw SQL when unmodeled (seamless to
   the user — "it's just magic").
 - Produce an **evaluation** comparing free-SQL vs via-Cube answers → an adoption verdict.
 - Build the **self-improving loop** that grows the model from usage (persist knowledge).
+- **Secondarily**, add synthetic connect-labs data as a reproducible alternate source (same
+  Connect export shape) once connect-labs#637 lands — primarily to strengthen the eval (§4.5).
+
+**Data-source priority.** Real Connect is the primary POC fuel; connect-labs synthetic is a
+follow-on. The auto-modeling, Cube, agent, and eval components are **data-source-agnostic** —
+they operate on whatever Connect data is materialized, so adding the synthetic source later is
+purely an additional loader/credential, not a redesign.
 
 **Non-goals**
 - Cube Cloud / hosted MCP / D3 (we use Cube Core in Docker).
 - Replacing raw-SQL access wholesale — Cube is additive; we migrate domains incrementally.
 - Exposing Cube or any modeling concept to end users. Users only ever chat.
-- Real (non-synthetic) Connect ingestion changes beyond what reuse naturally provides.
+- Blocking the POC on the synthetic connect-labs source.
 
 ## 3. End-state architecture
 
 ```
-connect-labs /api/export/  (incl. app_structure)            [external dep: connect-labs#637]
-      │   Scout Connect loader, re-pointed at synthetic base URL + PAT
+PRIMARY:  real CommCare Connect  (existing Connect pipeline + deliver-app CommCare app structure)
+SECONDARY: connect-labs /api/export/ (incl. app_structure)   [external dep: connect-labs#637]
+      │   Scout Connect loader (same export shape; synthetic = different base URL + PAT)
       ▼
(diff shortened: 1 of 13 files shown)


docs/superpowers/specs/2026-06-18-cube-semantic-layer-auto-modeling-design.md
```diff
@@ -0,0 +1,263 @@
...
+
+A `TransformationAssetRun`-style record tracks each generation run.
+
+### 4.3 Cube Core deployment (M3)
+
+- Add a `cube` service to `docker-compose` (and dev `Procfile`), reading `cube/model/`,
+  configured against `MANAGED_DATABASE_URL` + the synthetic schema (`search_path`).
+- Expose the **SQL API** (Postgres wire). Pre-aggregations optional — include a toggle so the
+  eval can measure the caching win.
+- Multi-tenancy: scope Cube to the workspace's schema (mirrors Scout's existing schema routing).
```
snopoke, line 179: This is the bit that concerns me. We have postgres schemas that are created dynamically. Do we need to deploy a cube instance per schema?

Also 'search_path' doesn't restrict access to a specific schema - it only sets it as the default. We'd need to address the access control issue and it's not clear to me from the docs how we'd accomplish that.

docs/superpowers/specs/2026-06-18-cube-semantic-layer-auto-modeling-design.md
```diff
@@ -0,0 +1,263 @@
...
+
+### 4.4 Agent ↔ Cube integration — interface 1, SQL API (M3)
+
+The 1-vs-2 interface choice is **invisible to users**; we pick the lowest-friction one.
+- New MCP tool **`semantic_query`** that connects to Cube's pg-wire SQL endpoint (parallel to
```
snopoke, line 184: If the agent is still writing raw SQL what's going to make it less prone to errors?

mcp_server/services/materializer.py
```diff
@@ -249,6 +252,30 @@ def _on_page(rows_loaded: int, rows_total: int | None) -> None:
...
+                    asset_result = upsert_connect_assets(tenant_membership.tenant, tenant_meta)
+                    logger.info(
+                        "Connect assets for %s: %d created, %d updated",
+                        tenant_membership.tenant.external_id,
+                        asset_result["created"],
+                        asset_result["updated"],
+                    )
+                    form_defs = (tenant_meta.metadata or {}).get("form_definitions", {})
+                    for ws in tenant_membership.tenant.workspaces.all():
+                        async_to_sync(sync_column_notes)(ws, "stg_visits", form_defs)
```
snopoke, line 272: This makes me thing that `TableKnowledge` shouldn't be tied to a workspace but rather a tenant. In fact I also don't know why `TenantMetadata` is linked to a `TenantMembership` and not directly to the `Tenant` (https://github.com/dimagi-rad/scout/issues/305)

snopoke (commented):
I haven't looked too closely at the code that converts the form def to tables and metadata but the concept seems good in general.

I also made some comments on the 'cube' design since that was in this PR as well.

docs/superpowers/specs/2026-06-18-cube-semantic-layer-auto-modeling-design.md
snopoke: Looks like https://github.com/dimagi-rad/scout/pull/303 has more context on this.

PR #303, opened 2026-06-18
Multi-tenant Cube.dev semantic layer: model generator + semantic_query tools + agent routing (M3)

jjackson:
## What this adds

A **multi-tenant Cube.dev semantic layer** over Scout's Connect data: the agent can query *governed metrics* (`MEASURE(...)`) instead of free-writing SQL, scoped correctly per workspace.

> **Stacked on #302** (Connect form-aware staging). This PR targets that branch and shows only the M3 commits; retarget to `main` once #302 merges.

### The architecture (Scout = single source of truth for tenancy)

1. Scout resolves `workspace_id → schema_name` via the existing `load_workspace_context()`, mints a short-lived **JWT** (HS256, `CUBEJS_API_SECRET`) carrying `securityContext = {workspace_id, schema_name}`, and connects to Cube's pg-wire SQL API with it.
2. Cube's `cube.js` verifies the JWT (`checkSqlAuth`), keys compilation/cache per workspace (`contextToAppId`), and serves each workspace its own model files (`repositoryFactory` → `cube/model/<schema_name>/`) — because different deliver apps have different columns.
3. Generated models use `sql_table: "{COMPILE_CONTEXT.security_context.schema_name}.stg_visits"`, so one model definition resolves to the correct per-tenant schema. Cube never re-implements Scout's routing.

### Components

- **`cube/cube.js` + `docker-compose.yml`** — multi-tenant Cube Core service (verified booting + JWT auth over pg-wire & REST against a live instance).
- **`apps/transformations/services/cube_model_{generator,schema}.py`** — LLM-driven generator (injectable client → deterministic tests) producing Pydantic-validated, multi-tenant Cube YAML; enforces `COMPILE_CONTEXT` in validation; one repair round-trip.
- **`mcp_server/services/semantic.py` + `server.py`** — `semantic_query` / `semantic_catalog` MCP tools (mint JWT, query Cube's SQL API / `/cubejs-api/v1/meta`); same envelope + server-side `workspace_id` injection as the raw `query` tool.
- **`apps/knowledge` `ModelGapSignal` + `apps/agents`** — agent prompt routing (prefer governed measures, fall back to raw SQL) + `record_fallback` primitive (detection wired in a later self-improving-loop phase).

## Security review

A dedicated whole-branch security review traced tenant isolation end-to-end and caught a **critical** defect (the two new tenant-data tools were missing from `MCP_TOOL_NAMES`, so server-side `workspace_id` injection didn't apply — an LLM/prompt-injection could supply another workspace's id and read its data). **Fixed + regression-tested** (`test_agent_graph_injection.py`): an attacker-supplied `workspace_id` is now overridden by the authoritative value, and both tools are asserted present in the injected set. JWT verification (signature + `exp` + pinned `HS256`), secret single-sourcing, and `COMPILE_CONTEXT` syntax all verified sound.

## Testing

~40 tests across the phase, all green: cube.js boot + JWT auth (live), model-generator contract + validation + repair, JWT mint/round-trip, `semantic_query`/`catalog` (mocked Cube), tenant-isolation injection regression, prompt routing, and `ModelGapSignal` persistence. No live LLM/Cube needed for the suite (injectable client + patched connections).

## Known follow-ups (documented, deliberately deferred)

- Point `CUBEJS_DB_USER` at a **read-only DB role** for defense-in-depth parity with the raw `query` path (Cube's SQL API is already read-only, so no active write path).
- Ensure `CUBEJS_DEV_MODE=false` in production (the dev compose enables the Playground).
- `record_fallback` is the persistence primitive; the *detection* of fallbacks is wired in the self-improving-loop phase.
- **Live end-to-end validation** (boot Cube + a real opportunity → generate model → agent answers via `semantic_query`) is the highest-signal next step before the evaluation phase.


cube/cube.js
```diff
@@ -0,0 +1,111 @@
...
+  contextToOrchestratorId: ({ securityContext }) =>
+    `SCOUT_${securityContext.schema_name}`,
+
+  /**
+   * Load per-workspace model files from cube/model/<schema_name>/.
+   *
+   * The Task 7 generator writes YAML files here after generating the model.
+   * If the directory is missing or empty, FileRepository returns an empty
+   * repository — Cube compiles an empty model, which is safe and returns
+   * an empty schema rather than erroring out.
```
snopoke, line 104: The generator lives in a different docker image - how does it write YAML files to the 'models' directory? Maybe it needs a shared docker volume or something?

apps/transformations/services/cube_model_generator.py
```diff
@@ -0,0 +1,403 @@
...
+# ---------------------------------------------------------------------------
+
+
+async def _call_model(client: Any, messages: list) -> str:
+    """Invoke model_client with messages; supports sync .invoke and async .ainvoke."""
+    if hasattr(client, "ainvoke"):
+        response = await client.ainvoke(messages)
+    else:
+        # Fallback: wrap sync invoke in a thread
+        response = await asyncio.to_thread(client.invoke, messages)
```
snopoke, line 248: this is unnecessary since we know the client has `ainvoke`

apps/transformations/services/cube_model_generator.py
```diff
@@ -0,0 +1,403 @@
...
+    # Support both LangChain message objects and plain strings/callables
```
snopoke, line 249: why not use structured output rather than manually parsing the output

apps/transformations/services/cube_model_schema.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @field_validator("sql_table")
+    @classmethod
+    def sql_table_has_compile_context(cls, v: str | None) -> str | None:
+        """Enforce that sql_table uses COMPILE_CONTEXT for multi-tenancy.
+
+        This is validated at the schema level so callers get a clear error
+        message when the multi-tenant template is missing.
+        """
+        if v is not None and "COMPILE_CONTEXT" not in v:
```
snopoke, line 186: This should be more explicit - it should be able to validate the exact value

apps/transformations/services/cube_model_generator.py
```diff
@@ -0,0 +1,403 @@
...
+
+def _parse_yaml(text: str) -> dict:
+    """Parse YAML text, stripping fences first. Returns parsed dict."""
+    cleaned = _strip_markdown_fences(text)
+    return yaml.safe_load(cleaned) or {}
+
+
+def _validate_cube_model(data: dict) -> CubeModel:
+    """Validate parsed YAML dict against CubeModel. Raises ValidationError on failure."""
+    return CubeModel.model_validate(data)
```
snopoke, line 203: does cube not have an API we could use to validate the structure?

apps/agents/services/fallback_log.py
```diff
@@ -0,0 +1,45 @@
...
+
+from typing import TYPE_CHECKING
+
+from apps.knowledge.models import ModelGapSignal
+
+if TYPE_CHECKING:
+    from apps.workspaces.models import Workspace
+
+
+async def record_fallback(
```
snopoke, line 24: I don't see any usages of this outside of tests

snopoke (commented):
This is interesting - having the LLM generate the cube files seem tricky to me since we'd different cube files across multiple runs which would break any features that are built on top of them e.g. artifacts.

Issue #254, opened 2026-06-15
[arch] Cost/latency floor: prompt caching, history + knowledge budgets, polling


bderenzi:
No Anthropic prompt caching anywhere; prune_messages is dead (history replayed unbounded, ~quadratic lifetime cost + eventual context overflow); knowledge injected with no budget; serial per-table TLS connections per cache miss; always-on polling; in-memory knowledge pagination; per-version artifact copies; OCS page_size; uncached me_view re-hitting all providers; DD N+1s.

## Findings (10)

- [ ] **`02#3`** DEBT/cost-perf (r5, mixed) -- System-prompt assembly opens serial per-table fresh TLS DB connections every 60s cache miss; no Anthropic prompt caching; per-process caches under 4 workers
- [ ] **`01#3`** LATENT/cost-perf (r5, accidental) -- prune_messages is dead code: conversation history replayed unbounded every LLM call, no trimming/summarization/tool-result compaction
- [ ] **`01#4`** LATENT/cost-perf (r7, accidental) -- Knowledge context injected into every system prompt with no size cap; retriever ignores its relevance argument
- [ ] **`13#1`** DEBT/cost-perf (r1, accidental) -- OCS loaders use upstream default page_size=100 — ~10-15x more list requests than necessary; ocs_participants docstring says 'no count today' but upstream now provides first-page count
- [ ] **`06#2`** DEBT/cost-perf (r3, accidental) -- Chat rate limiting uses per-process LocMemCache; ineffective across uvicorn workers and Redis is provisioned but unused
- [ ] **`05#6`** DEBT/cost-perf (r3, mixed) -- Always-on polling (jobs 3s + health 5s) with no visibility gating; each jobs poll triggers API-side janitor reconciliation (~5 DB queries)
- [ ] **`05#7`** DEBT/cost-perf (r2, accidental) -- Knowledge list endpoint paginates in memory over all rows; pagination has no UI so items beyond page 1 are unreachable
- [ ] **`09#9`** DEBT/cost-perf (r2, mixed) -- Artifact versioning copies full code per update and soft-delete never frees rows; live artifacts re-execute all source queries serially per open with no caching
- [ ] **`10#2`** DEBT/cost-perf (r1, accidental) -- Data dictionary view: per-table TableKnowledge N+1, two fresh managed-DB connections per request, and async_to_sync in a sync DRF view
[...]

PR #327, opened 2026-07-02
fix(deploy): resolve REDIS_URL in .kamal/secrets so the API/worker boot

bderenzi:
## Problem — the next domino after #326

With `MCP_SHARED_SECRET` fixed (#326, merged), the production deploy on `main` now gets **past Deploy MCP** ✅ and dies at **Deploy API**:

```
ERROR (Kamal::ConfigurationError): Secret 'REDIS_URL' not found in .kamal/secrets
```

Step conclusions for the merge-commit run confirm the cascade: `Deploy MCP: success`, `Deploy API: failure (REDIS_URL)`, Worker/Frontend skipped.

**Same root cause as #326:** wave-2 PR #321 (arch #254, shared Redis cache) added `REDIS_URL` to `config/deploy.yml`'s `secret:` block but never added a resolution line to `.kamal/secrets`.

## Fix

Define `REDIS_URL` from `SCOUT_REDIS_ENDPOINT` — the CloudFormation `RedisEndpoint` output, already a configured GitHub secret and already exported to the deploy env by `deploy.yml`:

```
REDIS_URL=redis://${SCOUT_REDIS_ENDPOINT}:6379/0
```

Per `infra/scout-stack.yml` the cluster is single-node, **no TLS** (`redis://`, not `rediss://`), **no auth**, port 6379 from EC2 only, and the endpoint is host-only. `config/settings/base.py` reads `REDIS_URL` optionally and connects lazily; if the endpoint is empty the app falls back to LocMemCache — so this can't regress container boot.

## Verified — this is the last gap

Ran a required-vs-defined audit of **all four** deploy configs' `secret:` blocks (`deploy.yml`, `deploy-mcp.yml`, `deploy-worker.yml`, `deploy-frontend.yml`) against `.kamal/secrets`. After this change the missing-secret set is **empty** — no third domino. `bash -n .kamal/secrets` parses.

## No operator action needed

Unlike #326, no new AWS secret is required — `SCOUT_REDIS_ENDPOINT` already exists as a repo secret and the ElastiCache cluster is already provisioned.

## Follow-up worth doing

This is the second time this exact bug shipped (MCP secret, then Redis). A cheap CI guard that parses the deploy configs' `secret:` blocks and asserts each name is defined in `.kamal/secrets` would catch the whole class. Happy to add it as a fast-follow.

https://claude.ai/code/session_01SxtpXTEsuK79vQHkTcJ2M5

### DIFF
--- a/.kamal/secrets
+++ b/.kamal/secrets
@@ -53,3 +53,12 @@ MCP_SHARED_SECRET=$(kamal secrets extract SCOUT_MCP_SHARED_SECRET $MCP_SECRETS)
 DATABASE_URL=$(scripts/resolve-database-url.sh)
 MANAGED_DATABASE_URL=$DATABASE_URL
 
+
+## Redis (ElastiCache) — shared cache for chat rate limiting / DRF throttles (arch #254).
+# Single-node cluster, no TLS, no auth (infra/scout-stack.yml: RedisSecurityGroup allows
+# 6379 from EC2 only). SCOUT_REDIS_ENDPOINT (CloudFormation RedisEndpoint output, exported
+# to the deploy env by deploy.yml) is the host address with no port. config/deploy.yml's
+# secret block lists REDIS_URL; without this line kamal errors "Secret 'REDIS_URL' not
+# found in .kamal/secrets". App falls back to LocMemCache if REDIS_URL is empty (base.py).
+REDIS_URL=redis://${SCOUT_REDIS_ENDPOINT}:6379/0
+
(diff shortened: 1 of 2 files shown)


bderenzi:
Added the guard test in this PR (commit `0c0761a`) rather than as a follow-up.

`tests/test_deploy_secrets.py` parses every `config/deploy*.yml` `env.secret:` block and `.kamal/secrets` off disk (no Kamal/AWS needed) and fails if any referenced secret is unresolved. It runs in the existing backend test gate, so from now on this class of bug blocks the **PR** instead of silently failing the **deploy**.

Verified locally:
- ✅ passes on this branch (REDIS_URL now resolved)
- ❌ fails with `Missing: {'deploy.yml': ['REDIS_URL']}` when the `REDIS_URL=` line is removed — i.e. it would have caught both this and #326 before merge.

snopoke (approved):

PR #328, opened 2026-07-02
feat: server-side tenant-access refresh + live enforcement (Root Cause A)

bderenzi:
## Root Cause A — server-side tenant-access refresh + live enforcement

Fixes the original report: a manager can't share a workspace with someone who *is* a Connect member ("User is not part of this workspace's tenants"), and access revoked upstream was never reflected in Scout. Scout previously learned who-belongs-to-what only at each user's own OAuth login, refreshed only the *sharer* (never the target), and gated access inconsistently.

Design doc: `docs/superpowers/specs/2026-06-18-tenant-access-refresh-design.md` (brainstormed + adversarially reviewed by a Fable subagent; every finding folded in before code).

### What changed (5 commits)

1. **Soft-delete manager on `TenantMembership`.** An archived membership is a *tombstone* for upstream-revoked access — not a human-recoverable soft-delete. Default `objects` is live-only (access reads safe by default); `all_objects` is the escape hatch for writes/merges/resolution. Merge gains **live-beats-tombstone** conflict semantics; API-key re-add reuses the tombstone. State-only migration.

2. **Provider-aware full-sync resolver.** Resolution now *archives* access the provider no longer returns (revocation), not just adds it. Mass-revoke footguns closed: CommCare pagination follows a relative `next` and **raises** on an off-host one (no silent truncation); **OCS archival is scoped to the token's team** (team-scoped tokens never revoke other teams); a 2xx body missing its expected key **raises** rather than being read as "zero tenants." Fail-open: any error/drift aborts before archival.

3. **One centralized authorizer** (`apps/workspaces/access.py`). Access = `WorkspaceMembership` **AND** (no tenants OR ≥1 *live* tenant). All gates delegate to it (workspace_resolver → every DRF/async endpoint; chat helper — closing the hole where a multi-tenant workspace granted access on membership alone). Dead `permissions.py` deleted. Unified gating: lose all upstream tenant access → lose the workspace (manage *and* query); regained automatically on upstream re-grant, never by a human in Scout.

4. **Share-time target refresh.** When adding a member who doesn't yet share a tenant, refresh the *target's* memberships server-side with their own **refresh-aware** token (bounded 8s) before failing — so someone granted access upstream after their last login is added without reconnecting. Error UX splits "reconnect" vs "no upstream access."

5. **CI fitness test** — fails the build if any module resolves workspace access by reading `WorkspaceMembership` filtered by both workspace and user outside `access.py`. Makes a future bypass un-mergeable.

### Tests
1353 pass. New coverage: manager/tombstone + merge conflict; full-sync add/un-archive/revoke, shape-drift, CommCare pagination, OCS team-scoping; authorizer + multi-tenant-hole closure; share-time refresh (grants / no-access / no-token); the fitness detector self-tests. Test fixtures updated to give members a live tenant (bulk_created to avoid the auto-workspace signal).

### Known residual (documented, follow-up)
**R1:** in a multi-tenant workspace the unioned view schema exposes all tenants' rows, so losing one tenant while keeping another still exposes the first's data. Matches today's add-member semantics; per-tenant query scoping is a follow-up. Also out of scope: C (invite/pre-share flow) and a periodic re-sync sweep.


tests/conftest.py
```diff
@@ -76,7 +81,14 @@ def other_user(db):
...
 def workspace(db, user, tenant):
-    """Create a test Workspace with WorkspaceTenant and WorkspaceMembership."""
+    """Create a test Workspace with WorkspaceTenant, WorkspaceMembership, and the
+    user's live TenantMembership — workspace access now requires a live tenant.
+
+    The TenantMembership is bulk_created so it does NOT fire the
+    auto_create_workspace_on_membership signal (which would spawn a second,
+    auto-created workspace and break count/listing assertions).
+    """
+    from apps.users.models import TenantMembership
```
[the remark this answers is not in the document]
bderenzi: Done in 53b445e — dropped all inline `TenantMembership` imports (the three you flagged plus the `tenant_membership` fixture's, which I was already editing). The module-level import at conftest.py:8 covers them.

apps/users/services/tenant_resolution.py
```diff
@@ -138,67 +181,72 @@ async def resolve_ocs_chatbots(user, access_token: str) -> list[TenantMembership
...
+            results.extend(data["objects"])
+            next_url = (data.get("meta") or {}).get("next")
+            if not next_url:
                 url = None
+            else:
+                resolved = urljoin(COMMCARE_DOMAIN_API, next_url)
+                if urlparse(resolved).netloc != base_netloc:
+                    raise TenantResolutionError(
+                        f"CommCare pagination points off-host, refusing to truncate: {next_url!r}"
+                    )
```
snopoke: This seems weird - why is it necessary?

snopoke (approved):

apps/users/services/tenant_resolution.py
bderenzi: Removed in 469e51e. The guard protected nothing: the same Bearer token is already sent to www.commcarehq.org on every page request, so a CommCare API compromised enough to hand back a malicious off-host `next` would already have the token. The part that actually mattered — resolving a *relative* `next` via `urljoin` so pagination isn't silently truncated after page 1 (which under full-sync would archive every domain past page 1) — stays; a malformed `next` still fails loud via `raise_for_status()`.

Issue #290, opened 2026-06-17
cancel_materialization writes RunState.FAILED on cancel (inconsistent with cancel endpoint's CANCELLED)


bderenzi:
Spun off from #289 (12#0 item 6).

The `cancel_materialization` MCP tool (`mcp_server/server.py`) marks a cancelled run as `MaterializationRun.RunState.FAILED` (and sets `result["cancelled"]=True`):

```python
previous_state = run.state
run.state = MaterializationRun.RunState.FAILED
run.completed_at = datetime.now(UTC)
run.result = {**(run.result or {}), "cancelled": True}
```

Meanwhile the user-facing cancel path (`apps/workspaces/api/jobs_cancel.py::cancel_thread_job`) writes the dedicated `RunState.CANCELLED`. The model has a distinct `CANCELLED` state, so a deliberately-cancelled run shows up as `FAILED` when cancelled via the MCP tool — inconsistent state semantics and confusing for any state-based reporting.

In #289 the hardened test pins the **current** behavior (asserts persisted `state==FAILED` + `result.cancelled==True`) so the seam is no longer invisible. This issue tracks the actual reconciliation: decide whether `cancel_materialization` should write `CANCELLED` (matching the endpoint) and update the pinned test accordingly.

Low priority / cleanup. Not blocking #289.

Issue #252, opened 2026-06-15
[arch] Credential lifetime for long jobs (CommCare 15-min OAuth TTL)


bderenzi:
Any CommCare-OAuth materialization >15 min is structurally impossible today: one credential snapshot per run, no mid-run/401 refresh, stale-token fallback, no 'reconnect your account' mapping. Plus CommCare/OCS retry hardening, uncancellable Retry-After sleep, and refresh-revokes-running-token race.

## Findings (7)

- [ ] **`14#3`** BROKEN-NOW/correctness (r1, accidental) -- One credential snapshot per materialization run with no mid-run/401 refresh: CommCare's 15-min OAuth TTL makes long CommCare OAuth syncs structurally impossible
- [ ] **`14#4`** BROKEN-NOW/correctness (r1, accidental) -- Refresh-failure falls back to a known-stale token and no 401 anywhere maps to re-authentication guidance; reactive-refresh path is dead code with false docstring
- [ ] **`12#4`** LATENT/correctness (r1, accidental) -- CommCare HQ actively rate-limits both APIs (429 + Retry-After by design); Scout's CommCare loaders have no retry and fail the run — raises the known Connect-only-hardening finding to expected-in-production for CommCare
- [ ] **`14#6`** LATENT/cost-perf (r1, accidental) -- Connect Retry-After honored uncapped inside the single worker thread; '~14s worst case' comment false and the sleep is uncancellable
- [ ] **`14#7`** LATENT/correctness (r1, accidental) -- Concurrent interactive token refresh (providers poll) can revoke the access token a long-running load is using mid-run
- [ ] **`03#6`** DEBT/correctness (r4, accidental) -- Retry/error-shape hardening applied only to Connect loaders; OCS/CommCare swallow malformed responses and have no retry
- [ ] **`09#3`** LATENT/correctness (r1, mixed) -- Mid-rematerialization reads: Connect serves silently-partial tables while CommCare/OCS block queries until 30s timeout

<details><summary>Evidence (claim / chain / files)</summary>

### `14#3` One credential snapshot per materialization run with no mid-run/401 refresh: CommCare's 15-min OAuth TTL makes long CommCare OAuth syncs structurally impossible
[...]

PR #332, opened 2026-07-03
arch #255: background-work robustness — reconcilers, cancel/pool/stream fixes, TTL & view-schema correctness

### LINKED ISSUE #255: [arch] Background-work robustness: concurrency, janitors, per-tenant locking
One worker at concurrency 1 serializes all background work; no janitor owns MaterializationRun after worker death (zombie 'doing' jobs); no per-tenant mutual exclusion; ThreadJob races its own dispatch; cancel semantics diverge; FutureApp current_app siblings; TTL rewind; checkpointer pool race; resume vs live turn unserialized; purge orphans; the dependency graph has no owner; stream 300s timeout only checked between events; job/checkpoint tables unpruned.

## Findings (16)

- [ ] **`08#2`** DEBT/cost-perf (r4, accidental) -- Single worker process at concurrency 1: all background work serialized platform-wide; worker deploy on every merge kills in-flight jobs
- [ ] **`03#9`** BROKEN-NOW/correctness (r4, accidental) -- No janitor reconciles MaterializationRun rows stuck in ACTIVE states after hard worker death; procrastinate stalled-job rescue never wired
- [ ] **`03#3`** LATENT/data-loss (r5, accidental) -- No per-tenant mutual exclusion for materialization: thread-scoped guard permits concurrent DROP/CREATE/INSERT on a shared tenant schema
- [ ] **`10#0`** LATENT/cost-perf (r3, accidental) -- Procrastinate job/event tables and LangGraph checkpoints never pruned; queue-table growth couples to janitor correctness
- [ ] **`01#2`** DEBT/correctness (r7, accidental) -- ThreadJob created after defer_async (dispatch/creation race); resume lifecycle is a long fix-chain held together by hedge + janitor
- [ ] **`01#1`** LATENT/correctness (r9, accidental) -- Two/three materialization-cancel mechanisms with diverging semantics; MCP cancel_materialization writes FAILED not CANCELLED and never stops the load
- [ ] **`07#1`** LATENT/correctness (r2, accidental) -- Cross-workspace cancellation: materialization_cancel_view selects orphan runs by shared tenant, cancels the same user's run in a sibling workspace
- [ ] **`02#5`** LATENT/correctness (r2, accidental) -- FutureApp import-time current_app binding (cause of the fixed janitor bug) survives at three sibling sites with swallowed exceptions
- [ ] **`04#0`** DEBT/correctness (r2, accidental) -- run_pipeline completion saves stale provision-time last_accessed_at, rewinding the TTL clock by the run duration
- [ ] **`03#1`** LATENT/correctness (r3, accidental) -- Partial-failure or cancel during multi-tenant rematerialization leaves the workspace's own view schema cascade-dropped but ACTIVE
- [ ] **`03#2`** LATENT/correctness (r2, accidental) -- build_view_schema reactivates EXPIRED view-schema rows without resetting last_accessed_at (incident-b class, view-schema edition)
- [ ] **`08#1`** LATENT/correctness (r2, accidental) -- Checkpointer pool singleton: unsynchronized init race; force_new closes the shared pool under concurrent streams; no borrow-time health check
- [ ] **`06#9`** LATENT/correctness (r1, accidental) -- No serialization between a resume ainvoke and a live user chat turn on the same LangGraph thread
- [ ] **`09#4`** DEBT/correctness (r1, accidental) -- purge_synced_data orphans WorkspaceViewSchema rows and ws_* physical schemas
- [ ] **`09#5`** DEBT/velocity (r3, accidental) -- Dependency graph between tenant schemas and view schemas has no owner: reconciled by 6+ hand-maintained hooks, four live mutation paths lack required hooks
- [ ] **`02#8`** LATENT/cost-perf (r3, accidental) -- Interactive chat stream 300s timeout only checked between events: a stalled LLM/tool call hangs the SSE response and leaks the abandoned generator

<details><summary>Evidence (claim / chain / files)</summary>

### `08#2` Single worker process at concurrency 1: all background work serialized platform-wide; worker deploy on every merge kills in-flight jobs
**Claim:** deploy-worker.yml runs bare 'procrastinate worker' with no --concurrency (default 1) and no queue, so all materializations, teardown/rebuild tasks, thread resumes, and both periodic janitors serialize through one slot — a long materialization blocks the TTL janitor and ThreadJob janitor entirely and delays every other workspace's work by its full duration (and inflates the created_at-anchored staleness that feeds the false-fail-resume race). tasks.py:368-370 relies on one-task-at-a-time for a timing argument, so the concurrency=1 assumption is load-bearing and unpinned. Every push to main redeploys the worker via Kamal with no stop_wait_time and procrastinate's shutdown_graceful_timeout defaults to None, so a running materialization is SIGKILLed after docker's grace — the deploy-triggered re-arm of the zombie-doing incident class.

**Files:** `config/deploy-worker.yml:16`, `.github/workflows/deploy.yml`, `apps/workspaces/tasks.py:368-370`, `.venv procrastinate/worker.py`

### `03#9` No janitor reconciles MaterializationRun rows stuck in ACTIVE states after hard worker death; procrastinate stalled-job rescue never wired
**Claim:** A worker that dies hard (SIGKILL after deploy stop-timeout, OOM, host crash) leaves the procrastinate job row 'doing' forever and the MaterializationRun in DISCOVERING/LOADING/TRANSFORMING forever: reconcile_stale_thread_job returns None for status in {todo,doing} (correct per-tick, permanent for zombies) and no code calls procrastinate 3.8's get_stalled_jobs/heartbeats/retry_job (grep: zero 'stalled' hits) and prune_stalled_workers never requeues jobs. No janitor iterates MaterializationRun rows at all. UI shows a frozen progress banner/spinner indefinitely (only escape is manual Stop then Retry); zombie rows block resume cursors from prior PARTIAL runs and are miscounted as active by the cancel endpoint. deploy-worker.yml has no stop_wait/shutdown-graceful-timeout and materializations routinely exceed docker's grace period, so every worker deploy mid-materialization re-arms this — the still-open 2026-05-30 zombie-doing incident class; June-10 fixed the dead-CONNECTION trigger, not the dead-PROCESS one. The docker-compose stack ships no worker at all (CLAUDE.md advertises 'all services'), so every materialization on the self-host compose path hangs forever.

**Files:** `apps/workspaces/tasks.py:728-838,748-749`, `apps/workspaces/api/jobs_views.py:129`, `config/deploy-worker.yml:16`, `docker-compose.yml`, `mcp_server/services/materializer.py:564-605,186-190`

### `03#3` No per-tenant mutual exclusion for materialization: thread-scoped guard permits concurrent DROP/CREATE/INSERT on a shared tenant schema
**Claim:** The in-flight dispatch guard filters ThreadJob by thread_id only; an in-code comment admits the materializer has no advisory lock per tenant_schema, so two threads in one workspace, two workspaces sharing a tenant, or a /materialize/retry/ racing a chat dispatch can run parallel run_pipeline against the same physical tenant schema. Interleaved DROP/CREATE/INSERT and per-page commits from two workers can deadlock or produce mixed table contents with both runs recording COMPLETED; the no-natural-PK resumable Connect tables can duplicate. The dead MATERIALIZING state (separate finding) removes the only single-tenant in-progress signal that might have deterred a second dispatch.

**Files:** `mcp_server/server.py:576-615`, `mcp_server/services/materializer.py:96-200`, `apps/workspaces/api/materialization_views.py:124-212`

### `10#0` Procrastinate job/event tables and LangGraph checkpoints never pruned; queue-table growth couples to janitor correctness
**Claim:** The worker runs without a job-retention flag and no remove_old_jobs/builtin maintenance task is scheduled, so procrastinate_jobs/procrastinate_events grow unboundedly (~144 janitor jobs/day plus every materialization/teardown/rebuild/resume). Conversely _procrastinate_job_status returns None for unknown ids and None means 'never touch the row', so if an operator ever prunes the table any still-active ThreadJob referencing a pruned id becomes permanently unreconcilable — retention and reconciliation are coupled but undesigned. The reconciler's status allowlist over raw strings also lets unknown/future procrastinate statuses fall into the act branch, and the dependency is pinned lower-bound-only.

**Files:** `config/deploy-worker.yml:16`, `apps/workspaces/tasks.py:693-725,748-816,819`, `pyproject.toml:34`

### `01#2` ThreadJob created after defer_async (dispatch/creation race); resume lifecycle is a long fix-chain held together by hedge + janitor
**Claim:** run_materialization (and the retry endpoint, duplicating the sequence) defers the procrastinate job before creating the ThreadJob with no shared transaction; the worker hedges with a bounded ~3.75s sleep-poll (_defer_resume_for_job) and otherwise leans on the 15-min janitor, giving a late-visible ThreadJob a phantom spinner up to ~10-25 min. tasks.py:373 TODO documents the clean fix (pre-create with nullable procrastinate_job_id; migration explicitly skipped). The resume/janitor lifecycle is the repo's largest fix-chain (~15-19 commits); invariants live only as CAS filter arguments and scattered comments rather than one owned module. If ThreadJob.acreate AND the rollback abort both fail (both swallowed), the materialization runs fully orphaned — expire_stale_thread_jobs iterates ThreadJobs only, so nothing reconciles it.

**Chain:** `mcp_server/server.py:606-635 defer_async then ThreadJob.acreate -> apps/workspaces/tasks.py:363-396 hedge + janitor fallback; TODO at tasks.py:373; duplicated at apps/workspaces/api/materialization_views.py:184-210`

**Files:** `apps/workspaces/tasks.py:363-396,690-838,1019-1289`, `mcp_server/server.py:520-648`, `apps/workspaces/api/materialization_views.py:184-210`, `apps/chat/models.py:75`

### `01#1` Two/three materialization-cancel mechanisms with diverging semantics; MCP cancel_materialization writes FAILED not CANCELLED and never stops the load
**Claim:** The API cancel view writes CANCELLED + procrastinate abort + ThreadJob flip; the worker's per-page cancellation checkpoint raises only on state==CANCELLED. The MCP cancel_materialization tool instead sets state=FAILED with result.cancelled=True, never aborts the procrastinate job and never touches the ThreadJob, so a mid-LOAD cancel does NOT stop loading — every page/source completes at full provider-API cost; the LOADING->TRANSFORMING CAS then misses and overwrites result, the run ends terminal FAILED with all sources committed, and downstream aggregation reports 'failed' not 'cancelled' (wrong resume message). Catalog only surfaces COMPLETED/PARTIAL runs, so the fully-loaded data presents as 'no data'. cancel_materialization and get_materialization_status also take an unscoped LLM-supplied run_id with no workspace/user binding (info disclosure: status leaks tenant external_id + physical schema name; cancel any in-flight run); HTTP equivalents scope to thread__user. Reviewers disagree whether these tools are in MCP_TOOL_NAMES/bound to the LLM; practical reachability is limited because no tool returns run_id post fire-and-ack.

**Chain:** `agent cancel_materialization(run_id) -> mcp_server/server.py:479 run.state=FAILED -> apps/workspaces/tasks.py:485-494 updater raises only on CANCELLED -> load continues -> materializer.py:435-443 CAS miss overwrites result, run FAILED with committed data -> metadata.py:57-67 hides it; contrast apps/workspaces/api/materialization_views.py:22-121 (CANCELLED + abort)`

**Files:** `mcp_server/server.py:407-493`, `apps/workspaces/tasks.py:485-494,991-1016`, `apps/workspaces/api/materialization_views.py:22-121`, `apps/agents/graph/base.py:65-76`, `mcp_server/services/materializer.py:430-443`

### `07#1` Cross-workspace cancellation: materialization_cancel_view selects orphan runs by shared tenant, cancels the same user's run in a sibling workspace
**Claim:** materialization_cancel_view selects active runs by tenant (tenant schemas are shared across workspaces) and matches ThreadJobs by thread__user only with no workspace filter, so cancelling in workspace A cancels the same user's chat-driven run started from workspace B sharing a tenant and injects a cancelled-resume into B's thread; the orphan path can cancel /refresh/-spawned runs of sibling workspaces. (Tracked jobs are correctly restricted to thread__user.)

**Files:** `apps/workspaces/api/materialization_views.py:46-117`

### `02#5` FutureApp import-time current_app binding (cause of the fixed janitor bug) survives at three sibling sites with swallowed exceptions
**Claim:** tasks.py documents and fixed (28b6647, ORM-based) the bug where module-level `from procrastinate.contrib.django.procrastinate_app import current_app` binds the unresolved FutureApp Blueprint (no job_manager). The identical import-by-value pattern survives at mcp_server/server.py:32, apps/workspaces/api/materialization_views.py:9, and apps/workspaces/api/jobs_cancel.py:11, all used for best-effort cancel/abort wrapped in try/suppress — they work only because views/MCP import after AppConfig.ready(); an import-order regression would silently disable all procrastinate aborts with no signal. The safe lazy ProxyApp spelling is already used by config/procrastinate.py.

**Files:** `apps/workspaces/tasks.py:693-725`, `mcp_server/server.py:32,630-635`, `apps/workspaces/api/materialization_views.py:9,105-118,200-208`, `apps/workspaces/api/jobs_cancel.py:11`

### `04#0` run_pipeline completion saves stale provision-time last_accessed_at, rewinding the TTL clock by the run duration
**Claim:** run_pipeline saves the TenantSchema at completion with update_fields=['state','last_accessed_at'] using the in-memory instance whose last_accessed_at was captured at provision time (run start), clobbering any newer touch made during the load and starting the 24h TTL clock H hours in the past for an H-hour pipeline — a smaller re-opening of the 2026-06-10 'janitor drops fresh schema' incident class for long loads plus quiet users.

**Files:** `mcp_server/services/materializer.py:485-486`, `apps/workspaces/services/schema_manager.py:114-122`, `config/settings/base.py:345`

### `03#1` Partial-failure or cancel during multi-tenant rematerialization leaves the workspace's own view schema cascade-dropped but ACTIVE
**Claim:** Per-tenant loads DROP TABLE ... CASCADE the tenant's raw tables, cascade-dropping the CURRENT workspace's namespaced ws_* views, but build_view_schema only runs when all_succeeded, and the sibling-rebuild explicitly excludes the current workspace. When one tenant fails or the user cancels mid-loop, the WorkspaceViewSchema stays ACTIVE missing the reloaded tenant's views; the resume's view_schema_failed check only fires when vs.state != ACTIVE, so the agent is told 'partial, answer what you can' / sources 'completed' while every query through the still-ACTIVE view schema hits missing views — the contradictory-catalog condition behind the #190 panic loops, whose canned 're-run materialization' advice happens to be the correct repair. PRs #227-#230 fixed this lie for sibling workspaces and the teardown path but missed the same-workspace partial/cancel path.

**Files:** `apps/workspaces/tasks.py:251-349,417-438,1063-1095`, `mcp_server/services/materializer.py:1418-1419`

### `03#2` build_view_schema reactivates EXPIRED view-schema rows without resetting last_accessed_at (incident-b class, view-schema edition)
**Claim:** The 2026-06-10 TTL fix reset last_accessed_at in provision() but not in build_view_schema, which get_or_creates the WorkspaceViewSchema row regardless of state and on success saves only update_fields=['state','last_error'] — a row resurrected from EXPIRED keeps its >24h-old last_accessed_at; teardown_view_schema_task also leaves EXPIRED rows with old timestamps and touch_workspace_schemas only touches ACTIVE/MATERIALIZING rows, so nothing refreshes the timestamp before/during rebuild. expire_inactive_schemas then flips the freshly rebuilt view schema to TEARDOWN on its next */30 tick (window narrowed by atouch on each MCP call). Bonus: the comment claiming build_view_schema saves state=FAILED before re-raising is false for early ValueErrors raised before the row is fetched — rows can park in PROVISIONING with no last_error.

**Files:** `apps/workspaces/services/schema_manager.py:243-447,280-287,436-438`, `apps/workspaces/tasks.py:547-606`

### `08#1` Checkpointer pool singleton: unsynchronized init race; force_new closes the shared pool under concurrent streams; no borrow-time health check
**Claim:** apps/chat/checkpointer.py builds a module-global AsyncConnectionPool + AsyncPostgresSaver with no lock: two concurrent cold-start requests interleave (the second closes the first's half-open pool, failing its setup); chat_view retries ANY build_agent_graph exception with ensure_checkpointer(force_new=True), which closes the pool that all other in-flight chat streams in that uvicorn worker are using for checkpoint writes — one request's transient error degrades every concurrent conversation; the pool is created without check= so dead pooled connections are handed out after a DB blip. AsyncPostgresSaver.setup() also runs at first use in every process — LangGraph checkpoint DDL applied at runtime outside Django migration discipline. With 4 workers x max_size 20, checkpointing alone can hold up to 80 platform-DB connections beside ORM and the worker, against db.t4g.micro's ~110 cap.

**Files:** `apps/chat/checkpointer.py:18,26-41`, `apps/chat/views.py:174-185`, `config/deploy.yml:17`, `infra/scout-stack.yml:16`

### `06#9` No serialization between a resume ainvoke and a live user chat turn on the same LangGraph thread
**Claim:** chat_view does not check for an active/RUNNING ThreadJob before streaming and the resume task does not check for an in-flight user turn; both invoke the graph against the same thread_id checkpoint concurrently. Dangling-tool-call repairs fix the Anthropic protocol symptom, but interleaved checkpoint writes from two concurrent writers can still interleave or drop superstep state. No CAS exists at this surface, unlike every other multi-writer row on the seam.

**Files:** `apps/chat/views.py:204-249`, `apps/workspaces/tasks.py:1143-1164`, `apps/chat/helpers.py:18-85`, `apps/agents/graph/base.py:551-578`

### `09#4` purge_synced_data orphans WorkspaceViewSchema rows and ws_* physical schemas
**Claim:** The purge_synced_data command tears down all tenant schemas and deletes TenantSchema rows but never touches WorkspaceViewSchema: physical ws_* schemas survive (emptied of views by CASCADE) and their rows keep their prior state (typically ACTIVE) over hollow schemas. Dev-targeted operator command.

**Files:** `apps/workspaces/management/commands/purge_synced_data.py`

### `09#5` Dependency graph between tenant schemas and view schemas has no owner: reconciled by 6+ hand-maintained hooks, four live mutation paths lack required hooks
**Claim:** The platform-side dependency graph (WorkspaceTenant) and physical-side graph (pg_depend via DROP CASCADE) are reconciled by at least six hand-maintained hooks scattered across materialize_workspace, teardown_schema, workspace_service, the resume task, chat touch, and MCP context; four live DDL-mutating paths lack required hooks (refresh, MCP teardown tool, transformations trigger, purge command) and the TTL keep-alive clause has five inconsistent implementations (only chat's touch_workspace_schemas is complete; MCP/recipes/artifacts/transformations under-touch, so a multi-tenant workspace kept alive only via non-chat surfaces expires its constituent schemas and the view schema flips FAILED while in active use). Each incident fix (#227-#230) added another hook at another site instead of a choke point — the contract is enforced by convention, already violated four times. When only the view schema expires (tenant data intact), the agent's only recovery is a full run_materialization re-sync because no tool/endpoint exposes the cheap rebuild_workspace_view_schema.

**Files:** `apps/workspaces/services/schema_manager.py`, `apps/workspaces/tasks.py`, `apps/workspaces/services/workspace_service.py:74-110`, `mcp_server/context.py:125`, `apps/chat/views.py:151`, `apps/recipes/api/views.py:113-122`

### `02#8` Interactive chat stream 300s timeout only checked between events: a stalled LLM/tool call hangs the SSE response and leaks the abandoned generator
**Claim:** langgraph_to_ui_stream computes a deadline but awaits event_stream.__anext__() with no timeout, evaluating the deadline only after an event arrives; a single LLM/tool call that stalls without emitting events makes the TimeoutError branch unreachable, so the 5-minute cap can be greatly exceeded and the response hangs until client disconnect (or a downstream SDK timeout). The resume path bounds the identical ainvoke with asyncio.wait_for; the interactive path did not (fixed-where-it-bit). On the timeout/exception branches the underlying async generator is abandoned without aclose(), so the in-flight ainvoke and its Anthropic call may keep running/billing until GC.

**Files:** `apps/chat/stream.py:39,84-260,100-112,212-239`, `apps/workspaces/tasks.py:1161-1164`

</details>

---
_Source: arch-review 2026-06-12 (repo HEAD 35e4230). Cluster `background-work-robustness`. See docs/arch-review/2026-06-12/synthesis.md._

bderenzi:
## Background-work robustness (arch #255)

Drives the concrete findings of the 16-finding `background-work-robustness` cluster to a reviewable state. Each fix is TDD'd and lands as its own commit. Design-gated and migration-risk findings are surfaced (with concrete proposals) rather than invented.

Verified current state first — several findings were already partly addressed by #317 (truthful cancel → CANCELLED), #319 (MCP cancel/status scoped to workspace), #311 (janitor logging), #328 (tenant-access). Only the residuals are fixed here; nothing prior is regressed.

### Fixed (concrete)

| id | fix |
|----|-----|
| **`02#5`** | Swap the three surviving `current_app` (FutureApp) sibling sites (`mcp_server/server.py`, `materialization_views.py`, `jobs_cancel.py`) to the lazy `app` (ProxyApp) from `config/procrastinate.py`, so an import-order regression can no longer silently disable procrastinate aborts. |
| **`01#1`** | MCP `cancel_materialization` now aborts the procrastinate job and flips the owning ThreadJob (it already wrote CANCELLED after #290, but left the job running and the chat spinner spinning). Mirrors the HTTP cancel path. |
| **`07#1`** | `materialization_cancel_view` scopes the tracked-cancel path by `thread__workspace` so cancelling from workspace A no longer cancels the same user's chat-driven run of a shared tenant in sibling workspace B (which injected a "cancelled" resume into B's thread). The orphan fallback stays workspace-agnostic on purpose — an orphan run has no ThreadJob, so cancelling it injects no resume into any thread. |
| **`04#0`** | `run_pipeline` completion now refreshes `last_accessed_at` to completion time instead of persisting the provision-time snapshot, so a long (H-hour) load no longer rewinds the 24h TTL clock by the run duration. |
| **`03#2`** | `build_view_schema` resets `last_accessed_at` on (re)build so a row resurrected from EXPIRED isn't immediately re-torn-down; the row is now created/reset **before** tenant validation so early "no tenants / tenant has no active schema" failures mark it FAILED (with a reason) instead of parking in PROVISIONING. Fixed the false comment in `rebuild_workspace_view_schema`. |
| **`03#1`** | Multi-tenant materialization now (re)builds the workspace's own view schema unconditionally (not only on full success), before the resume fires. A partial/cancelled run's tenant DROP cascade-dropped the ws_* views; leaving the schema ACTIVE made the resume tell the agent "answer what you can" over missing views. `build_view_schema` is idempotent and owns the FAILED marking. |
| **`08#1`** | Checkpointer pool: add an `asyncio` init lock (with double-check) so concurrent cold starts don't race; create the pool with `check=AsyncConnectionPool.check_connection` (borrow-time health check); and make `force_new` rebuild only the stateless saver over the still-open pool instead of closing a pool other in-flight streams are using. |
| **`02#8`** | `langgraph_to_ui_stream` wraps `event_stream.__anext__()` in `asyncio.wait_for(remaining)` so a stalled LLM/tool call that emits nothing still trips the 300s deadline, and `aclose()`s the generator in a `finally` so the abandoned ainvoke/Anthropic call is cancelled instead of leaking. |
| **`03#9`** | New periodic `reconcile_stale_materialization_runs`: detects a MaterializationRun stuck in an ACTIVE state after a hard worker death via procrastinate 3.8's heartbeat-based `get_stalled_jobs` (or a terminal job status), fails it truthfully, and fails its owning ThreadJob(s) with a synthetic chat message so the frozen UI banner clears. The ThreadJob janitor couldn't do this — it returns None for a `doing` zombie. |
| **`10#0`** | New daily `prune_old_procrastinate_jobs` (`delete_old_jobs`, 7-day horizon, succeeded-only) bounds `procrastinate_jobs`/`procrastinate_events` growth. Hardened `reconcile_stale_thread_job`'s status handling into explicit in-flight / failed / succeeded allowlists so an unknown/future status can never fall into the resume act branch (`aborting` now skips, `cancelled` now fails). The reconciler↔retention coupling is defused by the horizon being far longer than the 15-min stale-job window. |
| **`09#4`** | `purge_synced_data` now drops the ws_* view schemas and deletes the WorkspaceViewSchema rows, instead of orphaning them (ACTIVE rows over hollow schemas) after the tenant DROP CASCADE. |
| **`06#9`** | `chat_view` rejects a live turn with 409 when a RUNNING materialization ThreadJob owns the thread (a resume ainvoke is mid-flight against the same checkpoint — no CAS at this seam). See deferred note for the reverse guard. |

### Deferred / surfaced (not invented — concrete proposal below)

- **`03#3` — per-tenant mutual exclusion (design-gated).** No advisory lock keys materialization to a tenant schema, so two threads/workspaces sharing a tenant can run parallel `run_pipeline` against the same physical schema (interleaved DROP/CREATE/INSERT). **Proposal:** take a pg session advisory lock keyed on a hash of `tenant_schema` at dispatch (`pg_try_advisory_lock`), acquired around the load phase; on contention, either wait or short-circuit with an "already materializing this tenant" result. **Not implemented** here to avoid colliding with the #252 credential-lifetime work inside `materialize_workspace_core`.
- **`08#2` — single worker at concurrency 1 (design/infra).** All background work serializes through one slot and every deploy SIGKILLs in-flight jobs. Raising `--concurrency` interacts directly with `03#3` (per-tenant locking) and with `tasks.py`'s one-task-at-a-time timing assumptions, and needs a graceful-shutdown / `stop_wait_time` decision. Surfaced for infra design. (This PR's `03#9` reconciler makes the SIGKILL case self-heal, and `10#0` retention is the separable, safe part.)
- **`09#5` — dependency-graph choke-point owner (design).** The tenant↔view-schema dependency is reconciled by 6+ hand-maintained hooks with 4 unhooked mutation paths and 5 inconsistent TTL keep-alive implementations. A single owned reconciler/choke-point is a design change beyond this PR; surfaced. (This PR closes one same-workspace hole via `03#1` but does not centralize the contract.)
- **`01#2` — ThreadJob created after `defer_async` (migration risk).** The clean fix (pre-create ThreadJob with a nullable `procrastinate_job_id`, then patch it in) needs a migration the original TODO explicitly skipped. **Proposal:** make `procrastinate_job_id` nullable, create the ThreadJob PENDING before `defer_async` inside one `transaction.atomic()`, then `aupdate` the job id — removing the dispatch/creation race and the `_defer_resume_for_job` sleep-poll hedge. **Deferred** as medium-risk (schema migration + touches the resume lifecycle) to keep this PR reviewable; the existing hedge + janitor keep it correct today.
- **LangGraph checkpoint pruning (part of `10#0`).** Procrastinate job/event retention is implemented; pruning LangGraph checkpoints is a product-gated conversation-history retention decision (how long to keep chat history), so it's surfaced rather than chosen here.
- **`06#9` reverse guard.** The resume task detecting an in-flight live turn needs a live-turn marker that doesn't exist yet; the chat-side guard is implemented, the resume-side is surfaced.

### Cross-lane note
Shares `apps/workspaces/tasks.py` and `mcp_server/services/materializer.py` with the #252 (credential-lifetime) agent. This PR touches only its own territory (never the credential/OAuth threading, error-mapping funcs, or the `aresolve_credential` call site) and does not add per-tenant locking inside `materialize_workspace_core`, so the two branches should rebase cleanly.

### Tests
New/updated tests across `test_materialization_reconciler.py` (new), `test_checkpointer_pool.py` (new), `test_chat_stream_tool_cards.py`, `test_chat_thread_ownership.py`, `test_materialize_workspace_task.py`, `test_materializer.py`, `test_view_schema_builder.py`, `test_purge_synced_data.py`, `test_mcp_tenant_tools.py`, `test_jobs_endpoints.py`, `test_threadjob_janitor.py`. Full relevant suite (236 tests) green; `ruff check` + `ruff format` clean on all changed files.

Closes #255

### DIFF
--- a/apps/workspaces/management/commands/purge_synced_data.py
+++ b/apps/workspaces/management/commands/purge_synced_data.py
@@ -4,7 +4,12 @@ import logging
 
 from django.core.management.base import BaseCommand
 
-from apps.workspaces.models import TenantMetadata, TenantSchema, Workspace
+from apps.workspaces.models import (
+    TenantMetadata,
+    TenantSchema,
+    Workspace,
+    WorkspaceViewSchema,
+)
 from apps.workspaces.services.schema_manager import SchemaManager
 
 logger = logging.getLogger(__name__)
@@ -12,9 +17,10 @@ logger = logging.getLogger(__name__)
 
 class Command(BaseCommand):
     help = (
-        "Purge all materialized tenant data: drops managed-DB schemas, "
-        "deletes TenantSchema/MaterializationRun/TenantMetadata records, "
-        "and clears data dictionaries. Chat, artifacts, and learnings are preserved."
+        "Purge all materialized tenant data: drops managed-DB tenant AND view "
+        "schemas, deletes TenantSchema/MaterializationRun/TenantMetadata/"
+        "WorkspaceViewSchema records, and clears data dictionaries. Chat, "
+        "artifacts, and learnings are preserved."
     )
 
     def add_arguments(self, parser):
@@ -28,12 +34,14 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         schema_count = TenantSchema.objects.count()
         metadata_count = TenantMetadata.objects.count()
+        view_schema_count = WorkspaceViewSchema.objects.count()
         workspace_count = Workspace.objects.exclude(data_dictionary=None).count()
 
         self.stdout.write(
             self.style.WARNING(
                 f"\nPurge synced data summary:\n"
                 f"  TenantSchema records (+ cascaded MaterializationRun): {schema_count}\n"
+                f"  WorkspaceViewSchema records (+ ws_* schemas): {view_schema_count}\n"
                 f"  TenantMetadata records: {metadata_count}\n"
                 f"  Workspace records with data_dictionary to clear: {workspace_count}\n"
             )
@@ -46,6 +54,23 @@ class Command(BaseCommand):
         manager = SchemaManager()
         teardown_errors = []
 
+        # Drop the multi-tenant view schemas first. Their ws_* physical schemas are
+        # NOT reached by the tenant DROP ... CASCADE below (that only empties the
+        # namespaced views), so without this the ws_* schemas + WorkspaceViewSchema
+        # rows survive as orphans (rows typically still ACTIVE over hollow schemas)
+        # — arch #255, 09#4.
+        for view_schema in WorkspaceViewSchema.objects.all():
+            try:
+                manager.teardown_view_schema(view_schema)
+                self.stdout.write(f"  Dropped view schema: {view_schema.schema_name}")
+            except Exception as exc:
+                teardown_errors.append((view_schema.schema_name, str(exc)))
+                self.stdout.write(
+                    self.style.ERROR(
+                        f"  Failed to drop view schema {view_schema.schema_name}: {exc}"
+                    )
+                )
+
         for tenant_schema in TenantSchema.objects.all():
             try:
                 manager.teardown(tenant_schema)
@@ -56,6 +81,7 @@ class Command(BaseCommand):
                     self.style.ERROR(f"  Failed to drop schema {tenant_schema.schema_name}: {exc}")
                 )
 
+        deleted_view_schemas, _ = WorkspaceViewSchema.objects.all().delete()
         deleted_schemas, _ = TenantSchema.objects.all().delete()
         deleted_metadata, _ = TenantMetadata.objects.all().delete()
         Workspace.objects.update(data_dictionary=None, data_dictionary_generated_at=None)
@@ -64,6 +90,7 @@ class Command(BaseCommand):
             self.style.SUCCESS(
                 f"\nDone.\n"
                 f"  Deleted {deleted_schemas} TenantSchema/MaterializationRun rows\n"
+                f"  Deleted {deleted_view_schemas} WorkspaceViewSchema rows\n"
                 f"  Deleted {deleted_metadata} TenantMetadata rows\n"
                 f"  Cleared data_dictionary on {workspace_count} Workspace(s)\n"
             )
--- a/apps/chat/checkpointer.py
+++ b/apps/chat/checkpointer.py
@@ -1,5 +1,6 @@
 """Lazy singleton for the LangGraph async PostgreSQL checkpointer."""
 
+import asyncio
 import logging
 
 from django.conf import settings
@@ -13,45 +14,71 @@ logger = logging.getLogger(__name__)
 
 _checkpointer = None
 _pool = None
+# Serializes initialization so two concurrent cold-start requests can't interleave
+# (the second used to close the first's half-open pool, failing its setup — arch
+# #255, 08#1).
+_init_lock = asyncio.Lock()
+
+
+def _pool_is_usable(pool) -> bool:
+    """True if ``pool`` exists and has not been closed."""
+    return pool is not None and not getattr(pool, "closed", False)
 
 
 async def ensure_checkpointer(*, force_new: bool = False):
     global _checkpointer, _pool
+
     if _checkpointer is not None and not force_new:
         return _checkpointer
 
-    try:
-        database_url = get_database_url()
+    async with _init_lock:
+        # Re-check under the lock: another coroutine may have finished the build
+        # while we were waiting, in which case reuse it instead of rebuilding.
+        if _checkpointer is not None and not force_new:
+            return _checkpointer
 
-        if _pool is not None:
-            await _pool.close()
+        try:
+            database_url = get_database_url()
 
-        _pool = AsyncConnectionPool(
-            conninfo=database_url,
-            max_size=20,
-            open=False,
-            kwargs={
-                "autocommit": True,
-                "prepare_threshold": 0,
-            },
-        )
-        await _pool.open(wait=True, timeout=10)
+            # Reuse the existing pool if it is still open. force_new rebuilds the
+            # (cheap, stateless) saver but MUST NOT close a pool that other
+            # in-flight chat streams are borrowing from for checkpoint writes —
+            # closing it turned one request's transient error into a failure for
+            # every concurrent conversation in this worker (arch #255, 08#1). The
+            # borrow-time health check below recycles individually-dead
+            # connections, so a full pool rebuild is rarely needed anyway.
+            if not _pool_is_usable(_pool):
+                _pool = AsyncConnectionPool(
+                    conninfo=database_url,
+                    max_size=20,
+                    open=False,
+                    # Health-check each connection on checkout so a dead pooled
+                    # connection (after a DB blip) is recycled instead of handed
+                    # out mid-write.
+                    check=AsyncConnectionPool.check_connection,
+                    kwargs={
+                        "autocommit": True,
+                        "prepare_threshold": 0,
+                    },
+                )
+                await _pool.open(wait=True, timeout=10)
 
-        _checkpointer = AsyncPostgresSaver(_pool)
-        await _checkpointer.setup()
-        logger.info("PostgreSQL checkpointer initialized")
-    except Exception as e:
-        if settings.DEBUG:
-            logger.warning(
-                "PostgreSQL checkpointer unavailable, using MemorySaver (DEBUG only): %s", e
-            )
-            _checkpointer = MemorySaver()
-        else:
-            logger.error(
-                "PostgreSQL checkpointer failed in production — conversation history unavailable: %s",
-                e,
-                exc_info=True,
-            )
-            raise
+            _checkpointer = AsyncPostgresSaver(_pool)
+            await _checkpointer.setup()
+            logger.info("PostgreSQL checkpointer initialized")
+        except Exception as e:
+            if settings.DEBUG:
+                logger.warning(
+                    "PostgreSQL checkpointer unavailable, using MemorySaver (DEBUG only): %s", e
+                )
+                _checkpointer = MemorySaver()
+            else:
+                logger.error(
+                    "PostgreSQL checkpointer failed in production — conversation history "
+                    "unavailable: %s",
+                    e,
+                    exc_info=True,
+                )
+                raise
 
     return _checkpointer
--- a/apps/chat/checkpointer.py
+++ b/apps/chat/checkpointer.py
@@ -14,9 +14,7 @@ logger = logging.getLogger(__name__)
 
 _checkpointer = None
 _pool = None
-# Serializes initialization so two concurrent cold-start requests can't interleave
-# (the second used to close the first's half-open pool, failing its setup — arch
-# #255, 08#1).
+# Serialize init so concurrent cold starts don't race the half-open pool (arch #255 08#1).
 _init_lock = asyncio.Lock()
 
 
@@ -40,21 +38,14 @@ async def ensure_checkpointer(*, force_new: bool = False):
         try:
             database_url = get_database_url()
 
-            # Reuse the existing pool if it is still open. force_new rebuilds the
-            # (cheap, stateless) saver but MUST NOT close a pool that other
-            # in-flight chat streams are borrowing from for checkpoint writes —
-            # closing it turned one request's transient error into a failure for
-            # every concurrent conversation in this worker (arch #255, 08#1). The
-            # borrow-time health check below recycles individually-dead
-            # connections, so a full pool rebuild is rarely needed anyway.
+            # force_new rebuilds only the stateless saver; it must NOT close a pool
+            # other in-flight streams are still borrowing for writes (arch #255 08#1).
             if not _pool_is_usable(_pool):
                 _pool = AsyncConnectionPool(
                     conninfo=database_url,
                     max_size=20,
                     open=False,
-                    # Health-check each connection on checkout so a dead pooled
-                    # connection (after a DB blip) is recycled instead of handed
-                    # out mid-write.
+                    # Recycle a dead pooled connection on checkout, not mid-write (arch #255 08#1).
                     check=AsyncConnectionPool.check_connection,
                     kwargs={
                         "autocommit": True,
--- a/apps/chat/stream.py
+++ b/apps/chat/stream.py
@@ -25,6 +25,7 @@ Chunk types used:
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import hashlib
 import json
 import logging
@@ -207,12 +208,22 @@ async def langgraph_to_ui_stream(
         deadline = asyncio.get_event_loop().time() + AGENT_TIMEOUT_SECONDS
 
         while True:
-            if asyncio.get_event_loop().time() > deadline:
+            # Bound the wait for the NEXT event, not just the gap between events:
+            # a stalled LLM/tool call that emits nothing would otherwise never
+            # reach the deadline check (it fires only after an event arrives), so
+            # the 5-min cap could be blown wide open and the response hang until
+            # the client disconnects (arch #255, 02#8).
+            remaining = deadline - asyncio.get_event_loop().time()
+            if remaining <= 0:
                 raise TimeoutError(f"Agent execution exceeded {AGENT_TIMEOUT_SECONDS}s timeout")
             try:
-                event = await event_stream.__anext__()
+                event = await asyncio.wait_for(event_stream.__anext__(), timeout=remaining)
             except StopAsyncIteration:
                 break
+            except TimeoutError as exc:
+                raise TimeoutError(
+                    f"Agent execution exceeded {AGENT_TIMEOUT_SECONDS}s timeout"
+                ) from exc
 
             event_type = event.get("event")
 
@@ -461,6 +472,13 @@ async def langgraph_to_ui_stream(
                     "errorText": f"An error occurred while processing your request. Ref: {ref}",
                 }
             )
+    finally:
+        # Close the underlying event generator on every exit (timeout, error, or
+        # normal break) so a stalled/abandoned ainvoke — and the Anthropic call
+        # behind it — is cancelled promptly instead of running (and billing) until
+        # GC finalizes the orphaned generator (arch #255, 02#8).
+        with contextlib.suppress(Exception):
+            await event_stream.aclose()
 
     if reasoning_started:
         yield _sse({"type": "reasoning-end", "id": reasoning_id})
--- a/apps/chat/stream.py
+++ b/apps/chat/stream.py
@@ -208,11 +208,8 @@ async def langgraph_to_ui_stream(
         deadline = asyncio.get_event_loop().time() + AGENT_TIMEOUT_SECONDS
 
         while True:
-            # Bound the wait for the NEXT event, not just the gap between events:
-            # a stalled LLM/tool call that emits nothing would otherwise never
-            # reach the deadline check (it fires only after an event arrives), so
-            # the 5-min cap could be blown wide open and the response hang until
-            # the client disconnects (arch #255, 02#8).
+            # Bound the wait for the NEXT event: a silent stall (nothing emitted)
+            # would otherwise never reach the deadline check (arch #255 02#8).
             remaining = deadline - asyncio.get_event_loop().time()
             if remaining <= 0:
                 raise TimeoutError(f"Agent execution exceeded {AGENT_TIMEOUT_SECONDS}s timeout")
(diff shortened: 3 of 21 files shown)


apps/workspaces/management/commands/purge_synced_data.py
snopoke, line 1: when does this get used?

snopoke (approved):
The code comments are getting large again

apps/workspaces/management/commands/purge_synced_data.py
bderenzi: Dev-only reset command — run via `invoke purge_data` (or `manage.py purge_synced_data --confirm`); it's not wired into any automation or prod path. It wipes all materialized tenant data (tenant + view schemas, TenantSchema/MaterializationRun/TenantMetadata/WorkspaceViewSchema rows, data dictionaries) while preserving chat, artifacts, and learnings. The 09#4 change here just makes it also drop the ws_* view schemas so they don't survive as orphans after the tenant DROP…CASCADE.

bderenzi:
Addressed the review: trimmed the oversized comment blocks (the 4–12 line ones added across the PR) down to tight 1–3 line comments per CLAUDE.md in 3c45a85 — kept the non-obvious *why* + the `arch #255` refs, dropped the narration/history/restatement (net −61 comment lines). Also answered your inline question on `purge_synced_data`.

snopoke (approved):

PR #340, opened 2026-07-09
feat(workspaces): explain lost-upstream-access instead of a bare 403

bderenzi:
## What
#328's live-tenant gate returns a bare `403 "Workspace not found or access denied"` for two very different situations: (a) you were never a member, and (b) you're a member but lost all live upstream tenant access (e.g. a CommCare project was removed). Case (b) leaves users with a dead, unexplained workspace. In prod this affects 111 (user, workspace) pairs across 5 users.

## Change (does NOT weaken the gate)
**Backend** — `apps/workspaces/access.py`:
- New frozen `WorkspaceAccess` dataclass + `resolve_workspace_access_ex` / `aresolve_workspace_access_ex` that expose *why* access was denied (`not_member` vs `tenant_access_lost`, plus the lost tenant names).
- `access_denied_body()` builds the 403 payload: generic `{"error": ...}` for `not_member` (no info leak to non-members), and for `tenant_access_lost` an actionable message naming the project(s) + `reason` + `lost_tenants`.
- The existing `resolve_workspace_access` / `aresolve_workspace_access` **tuple API is preserved** as thin wrappers, so every untouched caller behaves identically. All gating stays in access.py (fitness test `test_authorizer_is_sole_gate.py` still passes).
- Wired through `workspace_resolver.py` (all DRF + async adapters) and the chat threads-load path.

**Frontend** — `ApiError` now carries the parsed body; `fetchThreads` surfaces the `tenant_access_lost` message in the sidebar (no Retry, since retry can't restore removed upstream access) instead of the generic "Couldn't load conversations".

## Tests
- New `tests/test_workspace_access_denial_reason.py` (sync + async, single/multi-tenant, granted + both denial reasons). Frontend `uiSlice.test.ts` covers both branches.
- Independently verified: `test_authorizer_is_sole_gate.py` + the new tests → **11 passed**. `ruff`, `bun run lint`, `tsc` clean.

## Reviewer notes / deferred follow-ups (NOT in this PR)
- ⚠️ Message wording is hardcoded to "CommCare project(s)" — slightly off for Connect/OCS-tenant workspaces; worth generalizing if multi-provider workspaces get common.
- Only the **threads-load** endpoint is enriched at the view level; other chat endpoints (`thread_messages_view`, `thread_share_view`, `apps/chat/views.py`) still emit the generic 403 because they route through `_resolve_workspace_and_membership`, which drops the reason. Threading it through that helper is a clean next step.
- Auto-reaping/hiding orphaned auto-created workspaces, and de-duplicating workspaces that share a display name after a CommCare domain slug change (e.g. `skelly` vs `skelly-1`), are separate tickets.

### DIFF
--- a/apps/workspaces/access.py
+++ b/apps/workspaces/access.py
@@ -63,7 +63,7 @@ def access_denied_body(result: WorkspaceAccess) -> dict:
         projects = ", ".join(result.lost_tenant_names)
         return {
             "error": (
-                f"You no longer have access to CommCare project(s): {projects}. "
+                f"You no longer have access to: {projects}. "
                 "Access may have been removed upstream — reconnect or ask an admin."
             ),
             "reason": TENANT_ACCESS_LOST,
--- a/apps/workspaces/access.py
+++ b/apps/workspaces/access.py
@@ -11,20 +11,86 @@ Because ``TenantMembership.objects`` is live-only (archived rows are tombstones 
 revoked access), the tenant check here naturally ignores revoked access. Every
 workspace-scoped view/tool MUST resolve access through this module; a CI fitness
 test (tests/test_authorizer_is_sole_gate.py) fails the build on a bypass.
+
+Denial is not one thing. A user who was never a member (or whose workspace is gone)
+gets a generic denial; a member who merely lost all live tenant access gets a
+distinct, actionable one naming the lost project(s) — so callers can explain "your
+upstream access was removed" instead of a dead, unexplained 403. The
+``(workspace, membership)`` tuple API is preserved; ``*_ex`` variants expose the
+reason, and ``access_denied_body`` builds the response payload from it.
 """
 
 from __future__ import annotations
 
+from dataclasses import dataclass
+
 from apps.users.models import TenantMembership
 from apps.workspaces.models import WorkspaceMembership
+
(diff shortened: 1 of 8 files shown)


apps/workspaces/access.py
```diff
@@ -11,20 +11,86 @@
...
+
+    Preserves the generic ``{"error": ...}`` shape for backward compatibility and,
+    for lost upstream access, adds ``reason`` + ``lost_tenants`` and an actionable
+    message the frontend can surface verbatim.
+    """
+    if result.denied_reason == TENANT_ACCESS_LOST and result.lost_tenant_names:
+        projects = ", ".join(result.lost_tenant_names)
+        return {
+            "error": (
+                f"You no longer have access to CommCare project(s): {projects}. "
```
snopoke: I don't think this error should be CommCare specific.

snopoke (commented):
LGTM pending the error message change

apps/workspaces/access.py
bderenzi: Good call — dropped the CommCare-specific wording in 8a6b624. The 403 message now just names the lost tenant(s) generically, which reads correctly for Connect/OCS workspaces too.

PR #341, opened 2026-07-10
feat(workspaces): intrusive lost-access UX + stop defaulting into dead workspaces

bderenzi:
## Why
#340 only put a small message in the chat sidebar when the active workspace's upstream access was removed. In practice the workspace was still fully clickable — every page just showed its own generic "Failed to load…" (each hitting a bare 403), and worse, `scout.dimagi.com` would *default* you into a dead workspace (e.g. `skelly`) because the client picks the first workspace the list API returns and that list includes orphaned memberships. This is the follow-up flagged in the #328/#340 investigation notes.

Two separate problems, both fixed here:
1. **UX** — you can wander a dead workspace with no clear explanation or way out.
2. **Routing** — a workspace you've lost access to shouldn't be your default landing spot or clutter the switcher (but should stay reachable by URL/bookmark for testing).

## Changes
**Backend — `apps/workspaces/api/workspace_views.py`**
- `GET /api/workspaces/` now returns `has_access` per workspace: `true` iff the workspace has no tenants or the user still shares a *live* `TenantMembership` with it — the same rule `apps/workspaces/access.py` gates each request with. One extra query for the whole list (tenants already prefetched). Rows are **annotated, not dropped**, so orphaned workspaces stay addressable by URL.

**Frontend — routing**
- Shared `workspaceHasAccess()` helper (missing flag ⇒ accessible, so a stale cached payload never locks the app out).
- `fetchDomains` defaults to the first *accessible* workspace instead of `domains[0]`.
- `WorkspaceSwitcher` lists only accessible workspaces in its segments / recent / search, and shows lost-access ones in a separate **disabled "No access" section** (visible for context, not selectable).

**Frontend — modal**
- New `LostAccessModal`, mounted in `AppLayout`: when the active workspace's `has_access === false` (stale default or deep link), a **non-dismissible** modal over a grayed backdrop. It names the provider (e.g. "This is a **CommCare** workspace"), explains access was removed upstream, points to the workspace owner / an admin, and offers a **picker of workspaces you can still access** — the only way out. Handles the no-accessible-workspaces case with guidance instead of an empty picker.

## Tests
- Backend: `tests/test_workspace_list_has_access.py` (live tenant / lost / zero-tenant / archived-tombstone). Broader workspace suite + `test_authorizer_is_sole_gate` still green (81 passed).
- Frontend: `LostAccessModal.test.tsx` (gating, provider naming, picker filtering, switch-on-click, no-access case) + `domainSlice.test.ts` default-pick cases. Full FE suite 73 passed; `tsc`/`bun run build`/`eslint`/`ruff` clean.

## Notes / not in this PR
- Visual QA on a real deploy still worth doing (I couldn't reproduce a lost-access workspace locally without prod-like data).
- Auto-reaping / de-duping orphaned auto-created workspaces (the `skelly` vs `skelly-1` slug/display collision) remains a separate ticket.

### DIFF
--- a/apps/workspaces/api/workspace_views.py
+++ b/apps/workspaces/api/workspace_views.py
@@ -236,6 +236,15 @@ class WorkspaceListView(APIView):
         memberships = list(memberships)
         schema_statuses = _schema_status_for_workspaces([m.workspace for m in memberships])
 
+        # Bulk live-access check, one query for the whole list. A workspace is
+        # accessible iff it has no tenants OR the user shares a live tenant with
+        # it — the same rule apps/workspaces/access.py enforces per request. We
+        # surface it here (rather than filtering rows out) so the client can keep
+        # orphaned workspaces addressable by URL while gating them in the UI.
+        user_live_tenant_ids = set(
+            TenantMembership.objects.filter(user=request.user).values_list("tenant_id", flat=True)
+        )
+
         results = []
         for m in memberships:
             tenants = [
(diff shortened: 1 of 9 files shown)


snopoke (approved):
It would be great to get a demo of what this looks like. Maybe [shot-scraper](https://shot-scraper.datasette.io/) could help.

PR #347, opened 2026-07-15
fix(workspaces): stop granting the view role access to raw tenant schemas

bderenzi:
Follow-up to #346. Removes cross-tenant over-exposure and a role-teardown leak in the workspace view-schema role.

## Problem

The workspace view role (`{view_schema}_ro`) was granted `USAGE` + `SELECT ON ALL TABLES` on every constituent raw tenant schema, plus (in `backfill_readonly_roles`) `ALTER DEFAULT PRIVILEGES` there. None of it is needed: workspace views are created with plain `CREATE VIEW` (no `security_invoker`), so reads through them resolve the underlying tenant tables with the **owner's** (CURRENT_USER's) privileges. The view role only needs access to the view schema.

Two issues with the old grants:

1. **Over-exposure** — a `SET ROLE {view_schema}_ro` session could read raw tenant tables directly, across tenant boundaries, instead of only through the workspace's namespaced views.
2. **Role-teardown leak** — the `ALTER DEFAULT PRIVILEGES` entries live in the tenant schemas (`pg_default_acl`), which outlive the view schema. `teardown_view_schema` drops only the view schema, so those entries survive `DROP SCHEMA CASCADE` and block `DROP ROLE`, leaving dangling roles (the error is best-effort-swallowed in teardown, so it accumulated silently).

## Changes

- `build_view_schema`: drop the per-tenant `USAGE`/`SELECT` grants; scope the stale-grant revoke to the view schema so a rebuild also strips grants left by older versions.
- `backfill_readonly_roles`: stop granting on tenant schemas; revoke stale grants instead, so running it remediates already-over-exposed roles.
- `_drop_readonly_role` / `_adrop_readonly_role`: also revoke `pg_default_acl` entries the role is a grantee of (keyed to the owning role) before `DROP ROLE`, so historical default-ACL leftovers no longer strand the role.

## Tests

- View role reads through the views but has no direct tenant access.
- Rebuild makes a removed tenant unreachable (no view + no grant).
- Backfill grants nothing on tenant schemas.
- Teardown drops the role despite a stale default-ACL entry (fails without the revoke with `DependentObjectsStillExist`).

## Remediation

Running `manage.py backfill_readonly_roles` on prod (already needed for #346) also revokes the stale over-exposing grants on existing view roles.

### DIFF
--- a/apps/workspaces/management/commands/backfill_readonly_roles.py
+++ b/apps/workspaces/management/commands/backfill_readonly_roles.py
@@ -102,39 +102,14 @@ class Command(BaseCommand):
         )
 
     def _backfill_view_schema(self, cursor, mgr, vs) -> None:
-        """Create the view-schema role and grant it access to constituent tenant schemas.
+        """Backfill the view-schema role.
 
-        Mirrors build_view_schema: the role gets SELECT on existing tables in each
-        constituent tenant schema AND default privileges, so a rematerialization
-        that runs between view rebuilds doesn't leave the new tables unreadable.
+        The view role needs access to the view schema ONLY: the views are owned by
+        CURRENT_USER and run with owner privileges, so the role never reads the raw
+        tenant schemas directly. Also revoke any tenant-schema grants left by
+        earlier versions — they widened cross-tenant reach and their default-ACL
+        entries blocked role teardown.
         """
         self._backfill_schema(cursor, mgr, vs.schema_name)
         role = readonly_role_name(vs.schema_name)
-        tenant_schemas_for_ws = TenantSchema.objects.filter(
-            tenant__in=vs.workspace.tenants.all(),
-            state=SchemaState.ACTIVE,
-        )
-        for ts in tenant_schemas_for_ws:
-            cursor.execute(
-                psycopg.sql.SQL("GRANT USAGE ON SCHEMA {} TO {}").format(
-                    psycopg.sql.Identifier(ts.schema_name),
-                    psycopg.sql.Identifier(role),
-                )
-            )
-            cursor.execute(
-                psycopg.sql.SQL("GRANT SELECT ON ALL TABLES IN SCHEMA {} TO {}").format(
-                    psycopg.sql.Identifier(ts.schema_name),
-                    psycopg.sql.Identifier(role),
-                )
-            )
-            # Default privileges so tables created later (by a rematerialization
-            # before the view is rebuilt) are still readable through this role.
-            cursor.execute(
-                psycopg.sql.SQL(
-                    "ALTER DEFAULT PRIVILEGES FOR ROLE CURRENT_USER IN SCHEMA {} "
-                    "GRANT SELECT ON TABLES TO {}"
-                ).format(
-                    psycopg.sql.Identifier(ts.schema_name),
-                    psycopg.sql.Identifier(role),
-                )
-            )
+        mgr._revoke_stale_view_role_grants(cursor, role, {vs.schema_name})
--- a/apps/workspaces/services/schema_manager.py
+++ b/apps/workspaces/services/schema_manager.py
@@ -403,11 +403,18 @@ class SchemaManager:
 
             view_role = readonly_role_name(view_schema_name)
 
-            # The _ro role is reused across rebuilds; revoke grants on tenants no
-            # longer in the workspace, else a removed tenant's data stays reachable
-            # (issue #244).
-            current_schemas = {view_schema_name} | {s for s, _ in tenant_schemas}
-            self._revoke_stale_view_role_grants(cursor, view_role, current_schemas)
+            # The view role needs access to the view schema ONLY. The views are
+            # owned by CURRENT_USER and created with plain CREATE VIEW (no
+            # security_invoker), so reads through them resolve the underlying
+            # tenant tables with the OWNER's privileges — the view role never
+            # touches the raw tenant schemas directly. Granting it USAGE/SELECT
+            # there was therefore unnecessary and widened cross-tenant reach (a
+            # SET ROLE {view_role} session could read raw tenant tables), and the
+            # tenant-schema default-ACL entries it left behind blocked DROP ROLE
+            # at teardown. Revoke any such grants left by earlier versions by
+            # scoping the "keep" set to the view schema alone (issue #244 cleanup
+            # of removed tenants still holds — they're revoked here too).
+            self._revoke_stale_view_role_grants(cursor, view_role, {view_schema_name})
 
             # ALTER DEFAULT PRIVILEGES only covers future tables, not the views
             # just created — grant SELECT on them explicitly.
@@ -417,21 +424,6 @@ class SchemaManager:
                     psycopg.sql.Identifier(view_role),
                 )
             )
-
-            # Views reference the constituent tenant schemas directly.
-            for tenant_schema_name, _ in tenant_schemas:
-                cursor.execute(
-                    psycopg.sql.SQL("GRANT USAGE ON SCHEMA {} TO {}").format(
-                        psycopg.sql.Identifier(tenant_schema_name),
-                        psycopg.sql.Identifier(view_role),
-                    )
-                )
-                cursor.execute(
-                    psycopg.sql.SQL("GRANT SELECT ON ALL TABLES IN SCHEMA {} TO {}").format(
-                        psycopg.sql.Identifier(tenant_schema_name),
-                        psycopg.sql.Identifier(view_role),
-                    )
-                )
 
             cursor.close()
         except Exception as exc:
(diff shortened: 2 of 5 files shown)


snopoke (commented):
I looked through the code but I wasn't able to do a detailed review

bderenzi:
## Triage: rebased onto `main`, both defects re-verified against real Postgres, one gap found and closed

This branch was ~7 weeks stale and had had no substantive review, so I re-verified the privilege claims empirically rather than from the Python. Summary: **the premise holds, Cube is unaffected, and the defect-2 fix as authored did not actually fix defect 2.** That last part is now fixed here.

### Rebase

Clean, no conflicts. #404 (the Cube semantic runtime merge) also touched `schema_manager.py`, but in `_create_readonly_role` — it added `GRANT SELECT ON ALL TABLES IN SCHEMA` for a schema's *own* `_ro` role. This branch touches `build_view_schema` and `_drop_readonly_role`. Both intents are intact post-rebase; #404's grant and its `ALTER DEFAULT PRIVILEGES FOR ROLE {dbt_role}` block are untouched.

### Defect 1 (over-exposure) — confirmed, and the fix is correct

Reproduced with a non-superuser owner role standing in for the managed DB user, over a tenant schema containing both a normally-owned table and a dbt-role-owned table.

Before (grants present), as `SET ROLE ws_..._ro`:

```
--- (3) DIRECT read of raw tenant A table ---      1 | TENANT-A-SECRET
--- (4) DIRECT read of raw tenant B table ---      1 | TENANT-B-SECRET   <-- cross-tenant
--- (5) DIRECT read of dbt-owned raw table ---     1 | DBT-OWNED-ROW
 usage_on_tenant_a | select_on_raw_a | select_on_raw_b
 t                 | t               | t
```

After this branch's change:

```
--- (1) read THROUGH the view ---                  1 | TENANT-A-SECRET   <-- still works
--- (2) read view over a DBT-OWNED table ---       1 | DBT-OWNED-ROW     <-- still works
--- (3) DIRECT read of raw tenant A table ---  ERROR: permission denied for schema t_pr347_a
--- (4) DIRECT read of raw tenant B table ---  ERROR: permission denied for schema t_pr347_b
--- (5) DIRECT read of dbt-owned raw table --- ERROR: permission denied for schema t_pr347_a
```

The load-bearing assumption — plain `CREATE VIEW` resolves the base tables with the **owner's** privileges — is confirmed, including for views over tables owned by the `_dbt` confinement role. `security_invoker` appears nowhere in the repo.

### Cube compatibility — not affected

This was the main risk, since #404 introduced a new DB consumer after this branch was written. It uses the same roles.

`cube_config/cube.js:60-72` builds its query driver from the JWT security context:

```js
options: `-c role=${readonlyRole} -c search_path=${schemaName},public -c statement_timeout=30000`,
```

`readonlyRole`/`schemaName` come from `build_cube_security_context` (`apps/semantic/services/cube_schema.py:225-233`) → `QueryContext.readonly_role` (`mcp_server/context.py:29-36`) → `readonly_role_name(schema_name)`. So it is exactly `{view_schema}_ro` for multi-tenant workspaces and `{tenant_schema}_ro` for single-tenant ones (`load_workspace_context`, `mcp_server/context.py:91-144`).

Crucially, the generated model never qualifies a schema (`apps/semantic/services/cube.py:62-65`, "Deliberately unqualified: the physical schema is resolved per query via the search_path"), and `search_path` is the routed schema plus `public` — so Cube reads **the views, not the raw tenant schemas**. Custom datasets go through `compile_custom_dataset_sql`, which rejects explicit schema qualification outright.

I simulated Cube's exact connection shape post-change:

```
$ PGOPTIONS='-c role=ws_..._ro -c search_path=ws_...,public -c statement_timeout=30000' psql
      current_user      |      current_setting
 ws_00000000000347aa_ro | ws_00000000000347aa,public
 SELECT * FROM "a__cases";       -> 1 | TENANT-A-SECRET
 SELECT * FROM "a__stg_visits";  -> 1 | DBT-OWNED-ROW        (view over a dbt-owned table)
 information_schema.columns for the view schema -> 6 rows   (driver introspection OK)
 SELECT * FROM t_pr347_a.cases;  -> ERROR: permission denied for schema t_pr347_a
```

Semantic queries work; the escape hatch is gone. Single-tenant workspaces route to `t_<id>` with `t_<id>_ro`, whose grants this branch does not touch at all.

Two things worth noting but out of scope here: Cube's control-plane pool (`cube.js:38`) runs as the `DATABASE_URL` owner with no role/search_path/timeout (two hardcoded parameterized `SELECT`s against `semantic_cubeschema`), and the healthcheck driver path (`cube.js:61-63`) skips the downgrade entirely, mitigated only by `repositoryFactory` returning no model files.

**Issue #406 is unrelated.** It is a semantic-layer capability gap — the agent can only aggregate over defined semantic members, so it declined free-text topic modelling over message bodies. It is not a Postgres privilege error and is not evidence that raw-schema access via the view role is intended. Nothing in it argues against this change.

### Defect 2 (teardown leak) — reproduced, but the fix here was incomplete

Reproduced on `main`:

```
DROP ROLE ws_..._ro;
ERROR:  role "ws_..._ro" cannot be dropped because some objects depend on it
DETAIL:  privileges for default privileges on new relations belonging to role pr347_owner in schema t_pr347_a
         privileges for default privileges on new relations belonging to role t_pr347_a_dbt in schema t_pr347_a
         privileges for default privileges on new relations belonging to role pr347_owner in schema t_pr347_b
         privileges for table t_pr347_a.later_table
```

Note that last line. The `pg_default_acl` revoke added here handles the first three, but not the fourth — and the fourth is reachable on the exact path this PR's own remediation creates:

1. `backfill_readonly_roles` (or a rebuild) revokes the schema-level `nspacl` grant in the tenant schema, but leaves the `pg_default_acl` entry.
2. The next materialization creates a table there. The surviving default ACL stamps `SELECT` for the view role onto it.
3. `_SCHEMAS_WITH_ROLE_GRANTS_SQL` only reads `pg_namespace.nspacl`, so with no schema-level entry left it returns **zero rows** — the table ACL is never found or revoked.
4. Teardown: `DROP ROLE` fails on `privileges for table ...` and `teardown_view_schema` swallows it. Same dangling role, same silence.

Verified by running `_drop_readonly_role` exactly as authored against that state:

```
-- step A: _SCHEMAS_WITH_ROLE_GRANTS_SQL (nspacl only) --   (0 rows)
-- step B: NEW _SCHEMAS_WITH_ROLE_DEFAULT_ACLS_SQL --       (3 rows, revoked fine)
-- step C: DROP ROLE --
ERROR:  role "ws_..._ro" cannot be dropped because some objects depend on it
DETAIL:  privileges for table t_pr347_a.later_table
=== survived? (1 row = STILL DANGLING) ===  ws_00000000000347aa_ro
```

The existing test passed only because it never creates a table *after* planting the stale default ACL.

### What I added (second commit)

- `_SCHEMAS_WITH_ROLE_GRANTS_SQL` now searches relation ACLs (`pg_class.relacl`) independently of schema ACLs, restricted to the relkinds `REVOKE ... ON ALL TABLES` can actually revoke. This fixes `_drop_readonly_role`, `_adrop_readonly_role`, and `_revoke_stale_view_role_grants` together, so a rebuild also strips table-level residue.
- `backfill_readonly_roles` now also calls a new `_revoke_stale_view_role_default_acls`, clearing the leftover `pg_default_acl` entries in the tenant schemas. Without this, remediation stops the direct grants but leaves the mechanism that keeps re-creating table ACLs. Deliberately kept out of `build_view_schema`: `ALTER DEFAULT PRIVILEGES FOR ROLE` requires membership in the owning role, and a failure there must not fail a rebuild.
- Two regression tests. The teardown one fails on this branch as originally authored (`AssertionError: view role ws_..._ro dangled after teardown`) and passes now.

### Tests

- Targeted (`test_view_schema_builder.py`, `test_schema_manager.py`, `test_backfill_readonly_roles.py`): **58 passed**.
- Full suite: **1876 passed, 5 skipped, 22 deselected** in 310s. Baseline on `main` @ `22889a7` is 1873/5/22, so +3 = the three new tests. No regressions.
- `uv run ruff check .` clean. Formatted only `tests/test_view_schema_builder.py` and `tests/test_backfill_readonly_roles.py` by explicit path (both were formatted on `main`); `schema_manager.py` and `test_schema_manager.py` are pre-existing unformatted on `main` and were left alone.

### Prod remediation still required

`manage.py backfill_readonly_roles` must run after merge. It is safe with the semantic runtime in place: it grants nothing new to view roles, revokes only their tenant-schema access, and Cube reads through `search_path`-resolved views. Existing view roles still hold their `nspacl` grants today, so the direct-ACL sweep finds them on the first pass; the default-ACL sweep then removes the re-accumulation source. Recommend `--dry-run` first — the command isolates per schema, so a single drifted row won't strand the rest.

One residual: view roles that already dangled in prod from before this fix have no `WorkspaceViewSchema` row to drive them, so neither the backfill nor teardown will reach them. Worth a one-off sweep for `rolname LIKE 'ws\_%\_ro'` with no matching row.

Issue #386, opened 2026-08-03
[tracking] Pair-debug 2026-08-03 remediation: build order and waves


bderenzi:
Build order for the 34 issues from the Brian/Simon multi-tenant workspace pair-debug session, 2026-08-03. Filter the batch with [`pair-debug-2026-08-03`](https://github.com/dimagi-rad/scout/issues?q=is%3Aissue+is%3Aopen+label%3Apair-debug-2026-08-03); each wave has a `pd-wave:N` label.

Waves are sized so one reviewer holds one mental model per wave. **Do not run them in parallel across agents** — waves 2 and 3 both touch `tenant_resolution.py` / `tasks.py`, and wave 0 changes the logging contract everything else is judged by.

---

## The single biggest finding: 7 issues are one missing abstraction

Scout has no notion of an **expected operational state** as distinct from a bug. So known conditions are raised as bare exceptions, `logger.exception`'d, and escalated to Sentry — where they crowd out real defects and page `#scout-ops`.

| Issue | Expected condition being escalated as an error | 90d events |
|---|---|---|
| #374 | dbt run failed (raw stdout shipped as error events) | 657 |
| #376 | LLM emitted a malformed artifact tool payload | 112 |
| #361 | a tenant hasn't been materialized yet | 57 |
| #371 | OAuth token expired or revoked | 69 |
| #372 | 401 vs 403 conflated, so the user gets advice that cannot work | subset |
| #373 | `invalid_grant` on a dead refresh token | — |
| #363 | state unclassified, so remediation advice is wrong | — |

**895 of 1,515 events (59%) are expected states.** #361 is the clearest illustration: `apps/workspaces/services/schema_manager.py:311` raises `ValueError` *after* `:320` has already correctly written `vs.state = FAILED` and `vs.last_error`. The state is modelled properly; the Sentry event is pure redundancy on top of it, and the condition is routine.

That is why **Wave 0 is the error taxonomy** — one primitive shrinks seven issues, and it makes Sentry trustworthy enough to tell whether later waves regress anything.

---

## Wave 0 — Error taxonomy `pd-wave:0`

Do this first. Small, low-risk, unblocks the rest.

- [ ] #371 Sentry has no expected-vs-unexpected classification (70% of last-24h events are routine OAuth expiry)
- [ ] #372 401 and 403 conflated in all three loaders
- [ ] #373 Token-lifecycle error classification
- [ ] #374 dbt duplicate fingerprints — 657 events, 43% of volume

[...]

Issue #374, opened 2026-08-03
dbt failures produce four duplicate Sentry fingerprints — 657 events, 43% of 90-day volume


bderenzi:
Found while quantifying Sentry noise during the Brian/Simon pair-debug session, 2026-08-03. Not observed in the session itself, but it is the single largest source of Sentry noise in the project and was surfaced by the same investigation.

## The numbers

Top 4 Scout Sentry issues by 90d event volume are all **the same dbt failure logged four different ways**:

| Rank | Short ID | Events | Title |
|---|---|---|---|
| 1 | `SCOUT-DJANGO-A` | 181 | `dbt run failed: dbt run failed` |
| 2 | `SCOUT-DJANGO-F` | 169 | raw dbt log `[error] … Encountered an error:` |
| 3 | `SCOUT-DJANGO-9` | 169 | raw dbt log `Encountered an error:` |
| 4 | `SCOUT-DJANGO-1T` | 138 | `TransformStageError: Stage 'system' failed: dbt run failed` |

**657 events = 43% of all 1,515 events in 90 days, for one root cause.**

The pattern repeats at lower volume: `2J`/`2G`/`2F`/`2E` (12 each), `2P`/`2N` (6 each), `2M`/`2K` (5 each), `2H`/`2D` (1 each).

## Cause

Raw dbt stdout is being shipped to Sentry as individual error events, *and* the wrapping exception is logged at multiple layers. Same double-log pattern as the auth-error issue, but with a third contribution: unstructured dbt log lines each fingerprinting separately.

## Why it matters

This is a bigger Sentry-hygiene win than the auth-error fix. It also actively hides real bugs — with 43% of volume being one duplicated failure, ranking by event count is misleading, and the `#scout-ops` Slack rule (`actionMatch: any`, no filters, fires on first-seen) pings on each new fingerprint.

## Suggested direction

- Stop forwarding raw dbt stdout to Sentry as error events; log it at WARNING or attach it as context/breadcrumbs on the single wrapping exception.
- Log the dbt failure **once**, at the boundary that has the most context, and let it propagate without re-logging.
- Consider an explicit `fingerprint` on the wrapping exception so all dbt-run failures group as one issue with a useful title.

Worth doing alongside the auth-error classification work — same mechanism (multi-layer `logger.exception`), same fix shape.

Issue #373, opened 2026-08-03
Token-lifecycle error classification: expected invalid_grant logged as an exception, and unknown expiry fails open


bderenzi:
Found in the Brian/Simon multi-tenant workspace pair-debug session, 2026-08-03. Two small, independent defects in the same module.

## Context from the session

Simon inspected the token record: *"token needs refresh to and expires_at within five minutes. So it says both have like a five minute…"*

For the record, those two facts are the *same fact viewed twice*, not a contradiction — `apps/users/services/token_refresh.py:27-28` sets `REFRESH_BUFFER = timedelta(minutes=5)` and `token_needs_refresh` derives the flag from `expires_at` with that buffer. What Simon saw was consistent.

## Defect 1: sync refresh logs expected failures as exceptions

`apps/users/services/token_refresh.py:140-142` — `refresh_oauth_token_sync` catches bare `Exception` and calls `logger.exception`, including for a `400 invalid_grant` (a dead refresh token), which is an **expected** outcome.

The async twin at `:88-99` explicitly gets this right, branching `400 <= status < 500` to `logger.warning` with the comment:

> "A 4xx (typically 400 invalid_grant on a dead refresh token) is an expected outcome, not a bug"

The sync (mid-run, reactive) path never received that treatment. **Cheapest single fix in the observability cluster** — mirror the async branch. Feeds the Sentry-noise issue.

## Defect 2: unknown expiry fails open, and reports as healthy

`token_refresh.py:52-60`:

```python
def token_needs_refresh(expires_at) -> bool:
    """...Returns False if expires_at is None (unknown expiry -- assume valid)."""
    if expires_at is None:
        return False
    return timezone.now() + REFRESH_BUFFER >= expires_at
```

A token with unknown expiry is never proactively refreshed and is handed straight to the loaders. Worse, it **reports as healthy** — `apps/users/auth_views.py:278-291` `providers_view` falls to `else: token_status[provider] = "connected"`, with no validation call anywhere. `expires_at` is the *only* signal.

So a revoked-but-unexpired token shows "connected" in the UI and is served to loaders, which then 401/403 — which is precisely the Sentry noise in the companion issue. **This is the mechanical link between the two.**

Note `expires_at` is only written when the provider returns `expires_in` (`:111-112`, `:148-149`), so a provider omitting it leaves the value at whatever it was — possibly `None`, possibly stale-past.

## Decision needed on defect 2

[...]

Issue #372, opened 2026-08-03
401 and 403 are conflated in all three loaders, so users are told to reconnect when that cannot help


bderenzi:
Found in the Brian/Simon multi-tenant workspace pair-debug session, 2026-08-03.

## The bug

`mcp_server/loaders/ocs_base.py:78-84`:

```python
if resp.status_code in (401, 403):
    raise OCSAuthError(
        f"OCS authentication failed for experiment {self.experiment_id} "
        f"(HTTP {resp.status_code}). Your Open Chat Studio sign-in has likely "
        f"expired or been revoked — please reconnect your account and retry."
    )
```

401 and 403 are merged into one exception with one message asserting the sign-in "has likely expired". Same pattern in `commcare_base.py:91` and `connect_base.py:127,195`.

**For a 403 that guidance is wrong.** With a valid token, OCS returns 403 when the token/team has no access to *that specific experiment* — access revoked upstream, the experiment moved teams, or the token is scoped to a different team (Scout already models team-scoping at `apps/users/services/credential_resolver.py:111-122`). Reconnecting mints an identically-scoped token and fails the same way, so **the user loops.**

The transport layer already draws the distinction the loaders then discard — `mcp_server/loaders/_http.py:76-96`:

> "It is consulted only on a 401 (an expiry signal); a 403 is a permission error left to the caller."

So a 403 is correctly never refresh-retried, and then incorrectly reported as an expiry.

## Production evidence

Both variants are live in Sentry: `SCOUT-DJANGO-2W`/`2V` (403, 2 events each, 2026-08-03) and `2X`/`2Y` (401, 5 events each, same day). The 403 group's title names one experiment UUID but its events span **two different tenants** (`514e2e67…` and `3980e719…`) — grouping is on stack+logger, not message, so the titles are arbitrary.

Note for anyone chasing strings: production does **not** emit "access token may have expired" (that phrasing is in `tenant_resolution.py`, which has zero recent events). The live message is the "…sign-in has likely expired or been revoked…" text above.

## Suggested direction

- **401** → credential dead → expected, user-recoverable → "reconnect your account".
- **403** → token valid, no access to this resource → expected, but **not** fixed by reconnecting → distinct message naming the specific chatbot/opportunity ("your OCS account no longer has access to *Dimagi-bot*"), and it should feed the revocation path rather than the reauth path.

[...]

PR #387, opened 2026-08-03
wave0(1/4): consolidate duplicate auth errors + add expected-state taxonomy

### LINKED ISSUE #371: Sentry has no expected-vs-unexpected classification: 70% of recent events are routine OAuth expiry
Found in the Brian/Simon multi-tenant workspace pair-debug session, 2026-08-03.

## Symptom

Mid-session Brian received a Sentry alert email: `SCOUT-DJANGO-2W — OCSAuthError: OCS authentication failed for experiment … (HTTP 403)`. Simon: *"I saw some that were like expired tokens that… I don't think they should be reported to Sentry probably. That should just be the — as long as there's user feedback, it's probably fine."*

## The numbers (from the Sentry API, project `dimagi/scout-django`)

| Window | Total error events | Auth-expiry events | **% auth** |
|---|---|---|---|
| **24h** | 20 | 14 | **70.0%** |
| **7d** | 21 | 14 | **66.7%** |
| 14d | 32 | 14 | 43.8% |
| 30d | 474 | 20 | 4.2% |
| 90d | 1,515 | 69 | 4.6% |

Right now **two-thirds of everything Scout reports to Sentry is expected, user-recoverable OAuth expiry.** (The 90d figure is diluted by a May–July dbt incident that has since stopped.) By issue count it's 8 of 86.

The full auth-error family, 90d: `SCOUT-DJANGO-Y` CommCareAuthError 401 (**46 events**, already rank #10 overall), `2Y`/`2X` OCSAuthError 401 (5 each), `1X`/`1W` OCSAuthError 403 (4 each), `2W`/`2V` OCSAuthError 403 (2 each), `12` CommCareAuthError 401 (1). Zero events for `ConnectAuthError`, `TokenRefreshError`, `CredentialResolutionError`.

## Root cause

`config/settings/base.py:289-298` is the entire Sentry config — dsn, environment, release, `traces_sample_rate`, `send_default_pii`. **No `before_send`, no `ignore_errors`, no `LoggingIntegration` override anywhere in the repo.** So the default `LoggingIntegration` captures every `ERROR`-level log as an event. (The `propagate: False` on the `apps`/`mcp_server` loggers in `production.py:74-89` does not help — `LoggingIntegration` hooks `Logger.callHandlers`.)

Sentry itself records these as `mechanism: {type: "logging", handled: true}` — its own metadata says these are handled logs, not crashes.

## Amplification: one failure → two events → two issue groups

The same exception is logged with `logger.exception` at two layers, so each mints its own Sentry group:

- `mcp_server/services/materializer.py:349-353` then re-raises at `:391`
- caller `apps/workspaces/tasks.py:373-375` (or `:225-227` on the refresh path)

Confirmed empirically — three pairs identical in type, message, culprit, file, function, and event count, differing **only in `logger`**, with timestamps matching to the second:

| `2W` (logger `apps.workspaces.tasks`) | `2V` (logger `mcp_server.services.materializer`) |
|---|---|
| 13:04:35Z `Materialization failed for tenant 514e2e67…` | 13:04:35Z `Source participants failed for schema t_514e2e67…` |
| 12:21:55Z `Materialization failed for tenant 3980e719…` | 12:21:55Z `Source participants failed for schema t_3980e719…` |

Exactly **2× on events**, and groups fan out further because 401 vs 403 and `_get` vs `_paginate` also split fingerprints. **Today: 7 real auth failures → 14 events → 4 distinct new issue groups.**

## It pages people

`sentry alert issues list dimagi/scout-django`: rule **"Notify #scout-ops via Slack"** (id `10003423453`), `actionMatch: any`, **no filters**, conditions include `FirstSeenEventCondition` ("a new issue is created"). Because each 401/403 variant mints a *new* group, today's 4 new OCSAuthError groups = **4 Slack pings to #scout-ops** for 2 distinct recoverable user problems. A second rule (`16945063`) emails IssueOwners on high-priority issues, and `2W` is `priority: high` — so it emailed too.

## Precedent to extend

The codebase already has the right pattern, just not applied to auth errors — `mcp_server/services/query.py:145-152`:

```python
if code in (VALIDATION_ERROR, QUERY_TIMEOUT):
    # Expected user/LLM query errors ... log as warnings without a traceback so they
    # don't get escalated into Sentry error events.
    logger.warning("Query error for tenant %s: %s", ctx.tenant_id, message)
else:
    logger.error(..., exc_info=True)
```

## Sibling sweep — every place an expected auth failure reaches Sentry

| Raise site | Escalation | Note |
|---|---|---|
| `apps/users/services/tenant_resolution.py:186` OCSAuthError | `apps/users/signals.py:139-141` `logger.exception` | deliberate — `signals.py:126-131` says "surface loudly so Sentry pages" |
| `tenant_resolution.py:136` ConnectAuthError | `signals.py:133-136` | same shape |
| `tenant_resolution.py:233` CommCareAuthError | `signals.py:143-146` | same shape |
| `mcp_server/loaders/ocs_base.py:79-84` | `materializer.py:349` + `tasks.py:373` | **the reported alert**, 2 events |
| `mcp_server/loaders/commcare_base.py:92-96` | same 2-hop chain | identical structure |
| `mcp_server/loaders/connect_base.py:128`, `:196-199` | same 2-hop chain | |
| `apps/users/services/token_refresh.py:140-142` | direct ERROR | see companion token-lifecycle issue |
| `mcp_server/loaders/_http.py:91` | `logger.warning(..., exc_info=True)` | **already correct — use as the reference pattern** |

**Already correctly downgraded** (precedent, don't change): `apps/users/views.py:98-118`, `apps/users/auth_views.py:78-80`, `apps/workspaces/api/workspace_views.py:76-82`, `apps/users/services/credential_resolver.py:208`. Note 4 of 5 callers of `resolve_ocs_chatbots` already log at WARNING; only the login signal logs at ERROR — that inconsistency is itself a finding.

**Legitimately Sentry-worthy, leave alone:** `apps/users/adapters.py:48-51` (`InvalidToken` on Fernet decrypt = key rotation / corruption).

## One precondition before downgrading

Simon's condition was "as long as there's user feedback". That holds on the **materialization** path (`apps/workspaces/tasks.py:82-87` `_looks_like_auth_failure` → `_REAUTH_GUIDANCE` in the chat message). It does **not** hold on the **login-signal** path (`signals.py:139-146`), which has no user-facing surface at all — a failed OCS resolution at login leaves the user with an empty data-sources page and only a Sentry event. Add feedback there before silencing it.

## Bonus bug found while verifying

Each auth class is **defined twice** as unrelated classes: `apps/users/services/tenant_resolution.py:44,48,52` and `mcp_server/loaders/{ocs_base.py:25, commcare_base.py:25, connect_base.py:59}`. An `except OCSAuthError` importing one will **not** catch the other. This plausibly explains why these reach `except Exception` rather than being handled. Worth consolidating into one module as part of this work.

## Suggested direction

Introduce an `ExpectedUpstreamError` base (or a `before_send` deny-list), consolidate the duplicate class definitions, apply the `query.py:145-152` pattern to the loader/resolver sites, and de-duplicate the two-layer log. Add a Sentry alert filter so recoverable auth states never page `#scout-ops`.

bderenzi:
## What & why

Wave 0 of #386, PR 1 of 4. Groundwork only — **this PR changes no behaviour and no Sentry event volume.** It lands the two things the other three wave-0 PRs build on:

1. **Each provider auth error is now defined exactly once.** `CommCareAuthError`, `ConnectAuthError` and `OCSAuthError` were each defined *twice* as unrelated classes — once in `apps/users/services/tenant_resolution.py:44,48,52` and once in the matching `mcp_server/loaders/*_base.py`. Same name, different class object, so an `except OCSAuthError` that imported one sailed straight past the other. That is plausibly why these fall through to `except Exception` instead of being handled specifically (bonus bug in #371).

2. **An `ExpectedStateError` primitive plus a Sentry `before_send`.** Scout has no notion of an *expected operational state* as distinct from a bug, which is the single missing abstraction behind 7 of the 34 pair-debug issues and 59% of Sentry's 90-day event volume (#386).

Closes #371 — the classification mechanism #371 says is missing lands here. The remaining reclassification work is tracked separately: #372 (loader 401/403), #373 (token lifecycle), #374 (dbt fingerprints).

### The bar for calling something "expected"

Raising `ExpectedStateError` silences an alert, so the rubric is written into `apps/common/errors.py` and is the thing I'd most like reviewed. All four must hold:

| | Test | |
|---|---|---|
| 1 | **Known** | the condition can be named in advance, not "something went wrong in here" |
| 2 | **Routine** | it happens in the normal operation of a *correct* system; no code change would prevent it |
| 3 | **Surfaced** | somebody who can act on it is told, through a channel that is not Sentry |
| 4 | **Resolved** | there's a defined next step, or it's a legitimate no-op |

(3) is load-bearing and the easiest to wave through. It is exactly why `apps/users/signals.py:139-146` is **not** being reclassified — a failed tenant resolution at login leaves the user on an empty data-sources page indistinguishable from "this account has no data", and the Sentry event is the only signal anything went wrong. An expected state nobody is told about is not expected; it's a silent failure.

**Nothing changes in this PR.** That is the claim being made, and it's pinned by `TestBeforeSend::test_provider_auth_errors_are_not_yet_classified`.

| Condition | Before | After this PR | Why |
|---|---|---|---|
| `OCSAuthError` / `CommCareAuthError` / `ConnectAuthError` (401 or 403) | ✅ reaches Sentry, 2 events + up to 2 new groups per failure | ✅ **unchanged** | classes moved, not reclassified — they still subclass `Exception`. Reclassifying is PR 2 (#371/#372) |
| `TokenRefreshError`, `CredentialResolutionError`, `TenantResolutionError` | ✅ reaches Sentry | ✅ **unchanged** | untouched here; token lifecycle is PR 3 (#373) |
| Raw dbt stdout (`stdout_log` / `file_log` loggers) | ✅ 340 events / 90d | ✅ **unchanged** | these carry no exception, so `before_send` can't see them by design — needs `ignore_logger`, PR 4 (#374) |
| Any exception subclassing `ExpectedStateError` | n/a | 🚫 **dropped** | nothing subclasses it yet, so this row is empty in practice |
| A bug raised *while handling* an expected state | ✅ | ✅ **kept** | `before_send` deliberately does not walk `__cause__`/`__context__` |
| A bare `logger.error` with no exception | ✅ | ✅ **kept** | nothing to classify |

Proposed Sentry-side rule changes are in "Notes for reviewers" — I have not touched any Sentry issue, rule, or alert.

The pattern: **an exception class defined in more than one module, so `except` on one misses the other.**

- **Grep used:** `grep -rn --include="*.py" -E "^class \w*(Auth|Export|Resolution|Refresh)Error" apps mcp_server config`

- **Sibling sites found & disposition:**
  - [x] `apps/users/services/tenant_resolution.py:44` `CommCareAuthError` — **fixed** (moved to `apps/common/errors.py`, imported back)
  - [x] `apps/users/services/tenant_resolution.py:48` `ConnectAuthError` — **fixed**
  - [x] `apps/users/services/tenant_resolution.py:52` `OCSAuthError` — **fixed**
  - [x] `mcp_server/loaders/commcare_base.py:25` `CommCareAuthError` — **fixed** (duplicate definition removed, now imports the shared class)
  - [x] `mcp_server/loaders/connect_base.py:59` `ConnectAuthError` — **fixed**
  - [x] `mcp_server/loaders/ocs_base.py:25` `OCSAuthError` — **fixed**
  - [ ] `mcp_server/loaders/commcare_base.py:29` `CommCareExportError` — **not applicable**: defined once, no duplicate. Left in place.
  - [ ] `mcp_server/loaders/connect_base.py:63` `ConnectExportError` — **not applicable**: defined once. Carries provider-specific structured fields (`status`, `attempts`, `sentry_trace`, `last_id`) that only the Connect loader populates; moving it would be churn.
  - [ ] `mcp_server/loaders/ocs_base.py:29` `OCSExportError` — **not applicable**: defined once.
  - [ ] `apps/users/services/tenant_resolution.py:45` `TenantResolutionError` — **not applicable**: defined once, and specific to tenant resolution (shape-drift guard). Left in place; it gets classified, not moved, in a later PR.
  - [ ] `apps/users/services/token_refresh.py:48` `TokenRefreshError` — **not applicable**: defined once. Touched by PR 3 (#373).
  - [ ] `apps/users/services/credential_resolver.py:41` `CredentialResolutionError` — **not applicable**: defined once.

Also swept for the inverse (a name imported from two places): every existing `from ... import *AuthError` in `apps/`, `mcp_server/` and `tests/` still resolves, and now resolves to the same object — asserted directly in `TestAuthErrorConsolidation`.

## Testing

`tests/test_error_taxonomy.py`, 15 tests:

- **Consolidation** — asserts class *identity* (`resolution_cls is loader_cls`), not just that both names resolve; two classes sharing a name is the bug. Plus one behavioural test: raising via the resolver import path is caught by `except` on the loader import path.
- **`before_send`** — both directions: drops `ExpectedStateError`/`ExpectedUpstreamError`; keeps unclassified exceptions, keeps events with no exception, and keeps a `ValueError` raised `from` an `ExpectedStateError` (the no-chain-walking rule).
- **Scope pin** — the three provider auth errors are asserted *not* to be dropped, which is this PR's "no behaviour change" claim. PR 2 inverts that assertion.

Full backend suite passes (`uv run pytest`). The 142 tests specifically touching these classes and loaders pass unchanged — no test needed editing, which is the point of importing the names back into their old modules.

## Notes for reviewers

- **Import direction.** `mcp_server/loaders/*` now imports from `apps.common.errors`. Precedent: `mcp_server/context.py:15` already imports `apps.common.identifiers`. `apps/common/errors.py` imports nothing, so there's no cycle and no Django-setup ordering hazard (`apps.common` is not an installed app).
- **`config/sentry.py` is imported from `config/settings/base.py`,** i.e. before `django.setup()`. It only imports the exception module, so this is safe — but it's worth a look.
- **`ExpectedUpstreamError` is currently unused.** Deliberate: it lands here so PR 2 is a reparenting diff rather than a new-abstraction diff. If you'd rather it arrive with its first user, say so and I'll move it into PR 2.
- **Proposed Sentry-side rule changes (not applied — Sentry is read-only for this work).** These are complementary to `before_send`, not redundant: `before_send` stops the events, the rule changes stop the *paging* for anything that slips through.
  - Rule "Notify #scout-ops via Slack" (`10003423453`) is `actionMatch: any` with **no filters** and includes `FirstSeenEventCondition`. Suggest adding a filter so it fires only on `level:error` **and** issues not tagged `expected_state`, or dropping `FirstSeenEventCondition` in favour of a frequency threshold — today 2 recoverable user problems produced 4 new groups and 4 Slack pings.
  - Rule `16945063` emails IssueOwners on high-priority issues; `SCOUT-DJANGO-2W` is `priority: high`, so it emailed too.
- **Follow-up, not in wave 0:** the login-signal path (`signals.py:139-146`) needs user-facing feedback before it can ever be downgraded. Cheapest honest option looks like a new `needs_reconnect` value on the `status` field `providers_view` already returns (`apps/users/auth_views.py:266-315`) — the frontend already renders `"expired"` as an amber "Connection expired" + "Reconnect" CTA (`ConnectionsPage.tsx:200-243`), so it's ~30 lines and no migration. Worth its own issue. Note that path currently emits **zero** Sentry events in 90 days (verified: `logger:apps.users.signals` returns nothing), so there is no noise pressure forcing it.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,6 +70,9 @@ SENTRY_DSN=
 # SENTRY_RELEASE=
 # SENTRY_TRACES_SAMPLE_RATE=0.0
 # SENTRY_SEND_DEFAULT_PII=False
+# Drop known operational states (apps/common/errors.py ExpectedStateError) before
+# they reach Sentry. Set to False to report everything again during an incident.
+# SENTRY_SUPPRESS_EXPECTED_STATES=True
 
 # Task Badger background-task tracking (optional — leave TASKBADGER_API_KEY blank to disable)
 # Use a project API key (base64-encoded org/project/key).
--- a/.env.example
+++ b/.env.example
@@ -71,7 +71,7 @@ SENTRY_DSN=
 # SENTRY_TRACES_SAMPLE_RATE=0.0
 # SENTRY_SEND_DEFAULT_PII=False
 # Drop known operational states (apps/common/errors.py ExpectedStateError) before
-# they reach Sentry. Set to False to report everything again during an incident.
+# they reach Sentry. Set to False to report them again.
 # SENTRY_SUPPRESS_EXPECTED_STATES=True
 
 # Task Badger background-task tracking (optional — leave TASKBADGER_API_KEY blank to disable)
--- a/apps/common/errors.py
+++ b/apps/common/errors.py
@@ -58,9 +58,7 @@ class ExpectedUpstreamError(ExpectedStateError):
     provider: str | None = None
 
 
-# --- Provider auth errors -------------------------------------------------
-#
-# These were each defined TWICE as unrelated classes — once in
+# These provider auth errors were each defined TWICE as unrelated classes — once in
 # ``apps/users/services/tenant_resolution.py`` and once in the matching
 # ``mcp_server/loaders/*_base.py`` — so an ``except OCSAuthError`` that imported
 # one sailed straight past the other (#371). They are defined here once and
--- a/apps/common/errors.py
+++ b/apps/common/errors.py
@@ -1,16 +1,7 @@
 """Scout's error taxonomy: expected operational states vs. defects.
 
-Scout had no way to say *"this failure is a known state, not a bug."* Every
-condition — a revoked OAuth token, a tenant that has not been materialized yet,
-an LLM emitting a malformed payload — was raised as a bare exception,
-``logger.exception``'d, and escalated into Sentry. Measured over 90 days, 59% of
-Sentry's event volume was routine states (#386). They crowd out real defects and
-page ``#scout-ops`` on every new fingerprint.
-
-``ExpectedStateError`` is the missing distinction. Anything raised as one is
-dropped by ``config.sentry.before_send``. It is still *logged* — the record
-survives in CloudWatch at WARNING — it just stops minting Sentry issues and
-paging people.
+An ``ExpectedStateError`` is dropped by ``config.sentry.before_send``. It is
+still logged — it just stops minting Sentry issues and paging people.
 
 Because raising one silences an alert, the bar is deliberately explicit. Raise
 ``ExpectedStateError`` only when **all four** hold:
@@ -24,14 +15,14 @@ Because raising one silences an alert, the bar is deliberately explicit. Raise
 4. **Resolved** — there is a defined next step (the user reconnects, the run
    retries), or the condition is a legitimate no-op.
 
-(3) is the load-bearing one and the easiest to wave through. It is why the login
-signal path (``apps/users/signals.py``) is deliberately *not* classified as
-expected: a failed tenant resolution there leaves the user with an empty
-data-sources page indistinguishable from "this account has no data", and the
-Sentry event is the only signal that anything went wrong at all.
+(3) is the load-bearing one and the easiest to wave through — an expected state
+nobody is told about is a silent failure, not an expected state. It is why the
+login signal path (``apps/users/signals.py``) is deliberately not classified as
+expected: a failed tenant resolution there leaves the user on an empty
+data-sources page indistinguishable from "this account has no data", so the
+Sentry event is the only signal anything went wrong.
 
-If any of the four fail, raise a plain ``Exception``. **An expected state that
-nobody is told about is not an expected state — it is a silent failure.**
+If any of the four fail, raise a plain ``Exception``.
 """
 
 from __future__ import annotations
@@ -50,19 +41,16 @@ class ExpectedUpstreamError(ExpectedStateError):
     """An expected state whose cause is an upstream provider, not Scout.
 
     ``provider`` lets a handler or a log line name the system that said no
-    without re-parsing the message. Subclasses set it as a *class* attribute so
-    existing ``raise SomeAuthError("message")`` call sites keep working
-    unchanged.
+    without re-parsing the message. It is a *class* attribute so that
+    ``raise SomeAuthError("message")`` call sites need no constructor change.
     """
 
     provider: str | None = None
 
 
-# These provider auth errors were each defined TWICE as unrelated classes — once in
-# ``apps/users/services/tenant_resolution.py`` and once in the matching
-# ``mcp_server/loaders/*_base.py`` — so an ``except OCSAuthError`` that imported
-# one sailed straight past the other (#371). They are defined here once and
-# imported by both: the point is identity, not a shared name.
+# Both apps/users/services/tenant_resolution.py and mcp_server/loaders/*_base.py
+# raise these, and each must catch what the other raises — so they live here, as
+# one class per provider rather than a name per module.
 
 
 class CommCareAuthError(Exception):
--- a/apps/common/errors.py
+++ b/apps/common/errors.py
@@ -1,16 +1,70 @@
-"""Errors shared across Scout's provider integrations.
+"""Scout's error taxonomy: expected operational states vs. defects.
 
-Each provider auth error used to be defined **twice** as two unrelated classes —
-once in ``apps/users/services/tenant_resolution.py`` and once in the matching
-``mcp_server/loaders/*_base.py``. An ``except OCSAuthError`` that imported one
-sailed straight past the other, which is why these reach ``except Exception``
-rather than being handled (#371).
+Scout had no way to say *"this failure is a known state, not a bug."* Every
+condition — a revoked OAuth token, a tenant that has not been materialized yet,
+an LLM emitting a malformed payload — was raised as a bare exception,
+``logger.exception``'d, and escalated into Sentry. Measured over 90 days, 59% of
+Sentry's event volume was routine states (#386). They crowd out real defects and
+page ``#scout-ops`` on every new fingerprint.
 
-They are defined here once and imported by both. The point is *identity*, not a
-shared name.
+``ExpectedStateError`` is the missing distinction. Anything raised as one is
+dropped by ``config.sentry.before_send``. It is still *logged* — the record
+survives in CloudWatch at WARNING — it just stops minting Sentry issues and
+paging people.
+
+Because raising one silences an alert, the bar is deliberately explicit. Raise
+``ExpectedStateError`` only when **all four** hold:
+
+1. **Known** — the condition can be named in advance ("the refresh token is
+   dead"), not "something went wrong in here."
+2. **Routine** — it occurs in the normal operation of a *correct* system. No
+   change to Scout's code would prevent it.
+3. **Surfaced** — somebody who can act on it is told, through a channel that is
+   not Sentry.
+4. **Resolved** — there is a defined next step (the user reconnects, the run
+   retries), or the condition is a legitimate no-op.
+
+(3) is the load-bearing one and the easiest to wave through. It is why the login
+signal path (``apps/users/signals.py``) is deliberately *not* classified as
+expected: a failed tenant resolution there leaves the user with an empty
+data-sources page indistinguishable from "this account has no data", and the
+Sentry event is the only signal that anything went wrong at all.
+
+If any of the four fail, raise a plain ``Exception``. **An expected state that
+nobody is told about is not an expected state — it is a silent failure.**
 """
 
 from __future__ import annotations
+
+
+class ExpectedStateError(Exception):
+    """A known, routine operational condition — not a defect.
+
+    Dropped from Sentry by ``config.sentry.before_send``. See the module
+    docstring for the four-part test a subclass must satisfy before it inherits
+    from this.
+    """
+
+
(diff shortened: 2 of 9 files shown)


bderenzi:
### Reading order — this is a 4-PR stack

Wave 0 of #386. Each PR is based on the previous one, so review in order and each diff is against its parent:

| | PR | Closes | Net Sentry effect |
|---|---|---|---|
| 1 | **#387** ← you are here | — | **none** (mechanism only) |
| 2 | #388 | #372, part of #371 | 61 auth events → 0; other materialization failures 2 events → 1 |
| 3 | #389 | #373 | ~0 (preventative; the win is correctness) |
| 4 | #390 | #374 | 662 → 138 |

Cumulatively ~39% of Scout's 90-day Sentry volume, and — the actual goal — event-count ranking becomes trustworthy again, which #386 says is a precondition for prioritising wave 1+.

Every PR body has a **before/after table** of exactly which conditions reach Sentry and which don't. That table is the thing to review; the code is mostly mechanical once you agree with the classification.

### CI note

`ci.yml` triggers only on `pull_request` with `branches: [main]`, so **#388/#389/#390 run `auto-review` but not lint/test** — they target stacked branches, not `main`. Each will get a full CI run when it's retargeted to `main` as its parent merges.

In the meantime I ran the full backend suite locally at the **tip of the stack** (all four PRs applied): **1772 passed, 5 skipped**, plus `ruff check` clean. Individual per-PR runs were green too.

### Not done, deliberately

- **No Sentry-side changes.** Sentry was read-only for this work — no issues resolved, muted, or edited, and no alert rules touched. Rule changes are *proposed* in the PR bodies instead. The main one: the `#scout-ops` rule (`10003423453`) is `actionMatch: any` with no filters and fires on first-seen, so any new fingerprint still pages even after this.
- **The login-signal path (`signals.py:139-146`) is untouched and still reaches Sentry.** #371's precondition says don't silence it without user-facing feedback, and #388 makes that structural rather than incidental — the classification lives on leaf classes the login path never raises, pinned by a test. That path emits **zero** events in 90d (`logger:apps.users.signals` → `"data": []`), so nothing forces it. Adding the feedback is a follow-up worth its own issue.
- **Wave 1+ not started.**

snopoke (approved):
The comments could be trimmed. They include motivation based on past repo state which is unncessary.

bderenzi:
Trimmed the comments per review — dropped the past-repo-state motivation (what the Sentry config used to lack, the duplicate-definition history, the 90-day event mix) from `apps/common/errors.py`, `config/sentry.py`, the settings comment and the test docstrings. What's left is the live rationale: the four-part bar for calling something expected, why the auth classes live in one module, and why `before_send` doesn't walk the exception chain. 6b3f26d.

Also added `Closes #371` to the body.

Issue #378, opened 2026-08-03
Upstream revocation never propagates: a 401/403 aborts tenant sync *before* the archival step


bderenzi:
Found in the Brian/Simon multi-tenant workspace pair-debug session, 2026-08-03. Related to #307 and #173, but this is the specific root cause neither states — see "Relationship to existing issues".

## Symptom (observed live on production)

Simon removed Brian from the `dimagi-dev` OCS team mid-session. Brian could still:

- continue the current chat and get answers over the materialized data,
- start a **new** chat in the same workspace and query it,
- log out and log back in (via a different provider) and still have workspace access,
- read artifacts,
- and — after Simon re-ran materialization from his own still-valid account — read **freshly loaded** data.

The only thing that broke was Brian's *own* "refresh data", which correctly 401'd.

## Root cause

`apps/users/services/tenant_resolution.py` full-syncs `TenantMembership` and archives rows that vanished upstream:

```python
archive_qs = TenantMembership.all_objects.filter(
    user=user, connection=connection, archived_at__isnull=True
).exclude(tenant_id__in=fresh_ids)
...
await archive_qs.aupdate(archived_at=timezone.now())
```

That block is at the **end** of the function. But an upstream 401/403 raises long before it:

- `tenant_resolution.py:136` — `ConnectAuthError`
- `tenant_resolution.py:186` — `OCSAuthError`
- `tenant_resolution.py:233` — `CommCareAuthError`

and every caller swallows the raise (`apps/users/signals.py:133-146`, `apps/users/views.py:115-118`, `apps/users/auth_views.py:76-79`). The module docstring states the policy outright:

> "Callers … treat any raise as 'skip refresh,' so access is never revoked on an inconclusive fetch (fail-open)."

**So the reaper only handles the narrow "token still valid, lost some tenants" case, and never the "removed from the account entirely" case — which is precisely the case that must revoke.** Total revocation *is* the 401/403.

Because `apps/workspaces/access.py` gates on a **live** `TenantMembership` and that row is never archived, the live-tenant gate is a no-op in the exact scenario it was built for.

## Production evidence

Queried the prod DB read-only, 2026-08-03, after Brian was removed from `dimagi-dev`:

```
-- Brian's OCS tenant memberships, grouped by team
team_slug     | total | live | archived
(none)        |    92 |   92 |        0
RAD-BMGF-FP   |    47 |   47 |        0
dimagi-dev    |    18 |   18 |        0
```

[...]

Issue #364, opened 2026-08-03
materialize_workspace silently skips tenants the triggering user lacks membership for, then reports success


bderenzi:
Found in the Brian/Simon multi-tenant workspace pair-debug session, 2026-08-03. Highest-severity finding in the materialization cluster.

## The bug

`apps/workspaces/tasks.py:285-300` resolves which tenants to load from the **triggering user's** memberships:

```python
qs = TenantMembership.objects.select_related("user", "tenant", "connection").filter(
    archived_at__isnull=True,
    tenant_id__in=[...the workspace's tenants...],
)
if user_id:
    qs = qs.filter(user_id=user_id)
memberships = [tm async for tm in qs]
```

`user_id` is always populated on the interactive paths (`mcp_server/server.py:724-727`, `apps/workspaces/api/materialization_views.py:161-164`). So in a shared workspace where member A added tenant 1 and member B added tenant 2, **any** materialization run covers only the triggering user's tenants. The others are not attempted, not reported, and not errored.

Then `tasks.py:377`:

```python
all_succeeded = all(r.get("success") for r in tenant_results)
```

A skipped tenant never enters `tenant_results`, so `all_succeeded` evaluates **`True`**. `_aggregate_materialization_state` (`:1356-1431`) likewise sees only the runs that happened and returns `status="completed"`.

**The run reports success while silently loading a subset of the workspace.** The only thing that catches it is the downstream view-schema failure — which then emits the misleading "a system-side fix is required" message (companion issue).

This is the mechanism behind the session's "it doesn't know about the other tenants for some reason".

## Zero test coverage on the production path

Every `materialize_workspace*` invocation in `tests/test_materialize_workspace_task.py` passes `user_id=""` (15 call sites). `test_run_materialization_fire_and_ack.py` passes real ids but only at the MCP-tool layer with the defer mocked.

So the `if user_id:` branch — **the only branch production uses** — has no coverage at all.

## Suggested direction

- Materialize the workspace's full tenant set, resolving a credential per tenant from **any** member who has one (which is already what the `user_id=""` path does), rather than restricting to the triggerer.
- If a tenant genuinely has no usable credential among the members, record it as a **failed/skipped** result so `all_succeeded` is False and the user is told which source could not be loaded and who needs to reconnect.
- Note the authorization tension: pulling with another member's credential is what the security cluster flags as a data-exposure vector. Whichever way this is resolved, the *silent* part must go — the run must not claim success for work it did not do.

## Tests to add

[...]

PR #388, opened 2026-08-03
wave0(2/4): split 401 from 403 and classify loader auth failures as expected

### LINKED ISSUE #372: 401 and 403 are conflated in all three loaders, so users are told to reconnect when that cannot help

bderenzi:
## What & why

Wave 0 of #386, PR 2 of 4. **Stacked on #387 — review that first; this PR's diff is against it.**

Closes #372. Part of #371 (the remaining half of #371 is PR 4's dbt noise).

Three things, one per commit:

1. **401 and 403 stop being the same thing.** All three loaders merged them into one exception asserting the sign-in "has likely expired… please reconnect your account". For a 403 that is wrong: the token is *valid* and simply has no access to that resource, so reconnecting mints an identically-scoped token and fails the same way — the user loops.
2. **The loader auth failures are classified as expected states,** so they stop paging. This is the two-line diff that matters; the rest is tests pinning what it means.
3. **One failure now produces one log record,** not one per layer.

### Verified against production Sentry (read-only)

Every auth group's culprit is a **loader**, not the login path:

| Group | Events (90d) | Logger | Culprit |
|---|---|---|---|
| `SCOUT-DJANGO-Y` | 46 | `apps.workspaces.tasks` | `commcare_base in _get` |
| `2Y` / `2X` | 5 / 5 | `materializer` / `tasks` | `ocs_base in _get` (401) |
| `2W` / `2V` | 2 / 2 | `tasks` / `materializer` | `ocs_base in _get` (403) |
| `12` | 1 | `apps.workspaces.tasks` | `commcare_base in _get` |

The `2Y`/`2X` and `2W`/`2V` pairs are the double-log: same culprit, same count, differing only in `logger`.

`sentry issue list --query "logger:apps.users.signals" --period 90d` returns **`"data": []`** — the login-signal path emits nothing at all. That fact drives the scope decision below.

|---|---|---|---|
| Loader 401 (`*TokenExpiredError`) | ✅ **2 events, 2 groups** per failure | 🚫 **dropped** | expected: the user gets `_REAUTH_GUIDANCE` in the chat failure summary |
| Loader 403 (`*AccessDeniedError`) | ✅ **2 events, 2 groups** per failure | 🚫 **dropped** | expected *only now* — its guidance became correct in commit 2 |
| **Login-signal auth failure** (`signals.py:139-146`) | ✅ reaches Sentry | ✅ **unchanged — still reaches Sentry** | **the hard constraint.** No user-facing surface exists, so it stays loud |
| `tenant_resolution` shape drift (`TenantResolutionError`) | ✅ | ✅ **unchanged** | not classified; aborts archival, genuinely unexpected |
| Any non-auth source failure (5xx, timeout, bad JSON) | ✅ **2 events, 2 groups** | ✅ **1 event, 1 group** | double-log removed; the inner WARNING becomes a breadcrumb |
| Materialization run failing before any source commits | ✅ 2 events | ✅ **1 event** | same |
| Asset-generation / transform / rollback failures | ✅ | ✅ **unchanged** | those handlers *swallow*, so their log is the only record — they stay ERROR |
| A bug raised while handling an auth error | ✅ | ✅ **kept** | `before_send` does not walk `__cause__` |
| Raw dbt stdout (340 events) | ✅ | ✅ **unchanged** | no exception attached; needs `ignore_logger` (PR 4) |

**Net: 61 auth events → 0, and every remaining materialization failure halves from 2 events to 1.** Estimated ~7 fewer new-issue Slack pings to `#scout-ops` per auth incident (today 2 user problems → 4 groups → 4 pings).

### The hard constraint, and why it needed a structural answer

The obvious implementation — reparent `OCSAuthError` under `ExpectedUpstreamError` — would have silenced `signals.py:139-146` **without anyone editing that file**, because `tenant_resolution` raises the same class. That's precisely the precondition #371 warns about: a failed resolution at login leaves the user on an empty data-sources page that reads as "this account has no data", and the Sentry event is the only signal anything broke.

So expectedness lives on the **leaf** classes, which only the loaders raise:

```
OCSAuthError(Exception)                     ← tenant_resolution raises this. NOT expected. Still pages.
├── OCSTokenExpiredError(…, UpstreamTokenExpired)   ← loader 401. Expected.
└── OCSAccessDeniedError(…, UpstreamAccessDenied)   ← loader 403. Expected.
```

`test_provider_base_classes_are_not_classified` pins it, so a future reparenting has to argue with a test. The reasoning is that rule 3 of the taxonomy ("someone who can act on it is told, through a channel that is not Sentry") is a property of the **raise site**, not the exception type — and the loader path satisfies it while the login path does not.

Giving the login path a real surface is a follow-up, not smuggled in here: a `needs_reconnect` value on the `status` field `providers_view` already returns, which `ConnectionsPage.tsx:200-243` already renders as an amber "Reconnect" CTA (~30 lines, no migration). Nothing forces it now — that path emits zero events.

### Two axes, and why both

`except OCSAuthError` still catches everything OCS. `except UpstreamAccessDenied` now catches every 403 across all three providers — the hook #378/#384 need, since a 403 is an authoritative per-tenant revocation signal.

The provider axis is load-bearing beyond `except`: `_summarize_error` serialises `exc.__class__.__name__` into `MaterializationRun.result`, so the class name is the part that survives the JSON round-trip into the user-facing summary. That's also why the guidance predicates match on class name rather than prose.

**Pattern A — `if resp.status_code in (401, 403)` conflating two different conditions.**

- **Grep used:** `grep -rn --include="*.py" "status_code in (401, 403)" apps mcp_server`

  - [x] `mcp_server/loaders/ocs_base.py:79` — **fixed** (split; 403 names the chatbot, does not say reconnect)
  - [x] `mcp_server/loaders/commcare_base.py:92` — **fixed**
  - [x] `mcp_server/loaders/connect_base.py:128` — **fixed** (routed through one `_raise_for_auth`)
  - [x] `mcp_server/loaders/connect_base.py:196` — **fixed** (the duplicate raise; same helper, so they cannot drift)
  - [x] `apps/users/services/tenant_resolution.py:124, :174, :221` — **partially fixed, deliberately.** Message corrected (no longer asserts expiry on a 403), but NOT split into typed classes: these are *collection* endpoints where a 403 means the token can't enumerate at all, and re-consenting against a different team genuinely may help. The loaders' "reconnecting will not help" copy would be wrong here. No type/control-flow change, so revocation safety is untouched.
  - [ ] `apps/users/services/api_key_providers/commcare.py:25` — **not applicable.** Raises `CredentialVerificationError` with `"CommCare rejected the API key (HTTP {status})"` — already status-accurate, gives no wrong advice, and runs synchronously while the user is looking at the form.
  - [ ] `apps/users/services/api_key_providers/ocs.py:38` — **not applicable**, same shape.

**Pattern B — `logger.exception` in a handler that re-raises (mints a second Sentry group).**

- **Grep used:** `grep -n "logger.exception" mcp_server/services/materializer.py`, then classified each by whether its handler re-raises.

  - [x] `materializer.py:350` per-source failure — **fixed** → WARNING (re-raises; `tasks.py` logs the ERROR)
  - [x] `materializer.py:429` pre-loop failure — **fixed** → WARNING (same)
  - [ ] `materializer.py:235` system-asset generation — **not applicable**: swallows ("continuing pipeline"), so this is the only record
  - [ ] `materializer.py:258` Connect-asset generation — **not applicable**: swallows
  - [ ] `materializer.py:466` transform phase — **not applicable**: swallows into `transform_result`. (Also PR 4 territory.)
  - [ ] `materializer.py:732` rollback failure — **not applicable**: logs a *different* exception (the failed rollback) than the one re-raised; a broken connection is its own defect
  - [ ] `apps/workspaces/tasks.py:335, :360, :374` — **not applicable**: these are the handlers that *stop* propagation. They are the single ERROR and must stay.

Verified there is no path where downgrading loses the only log: every live caller of `run_pipeline` logs at ERROR, and the one entry point that doesn't (`run_commcare_sync`) has zero callers.

## Testing

Full suite: **1751 passed, 5 skipped**. No existing test needed editing.

New:
- `tests/test_loader_401_403_split.py` (16) — all four raise sites × {401 → `*TokenExpiredError` and says reconnect; 403 → `*AccessDeniedError`, names the resource, says reconnecting will *not* help, and never says "reconnect your"}. Plus: a 403 is not refresh-retried, pinning that the transport-layer distinction is now preserved rather than discarded.
- `tests/test_single_log_per_failure.py` (3) — a **static AST check** over the whole materializer: no handler may call `logger.exception` and then re-raise. Written statically so a *newly added* offender fails the build rather than only the paths a test happens to exercise. The four legitimately-swallowing handlers are listed explicitly. **Verified non-vacuous** — reverted each site and confirmed both tests fail.
- `tests/test_error_taxonomy.py` (+15) — both inheritance axes, that the two causes don't catch each other, and the two `before_send` directions including the login-path guard.
- `tests/test_reauth_failure_mapping.py` (+4) — a 403 gets the access-removed guidance and *not* the reauth guidance; a run with one dead token and one revoked access carries both.

## Notes for reviewers

- **The judgment call to check is commit 3** (`feat(errors): classify loader auth failures…`). It's ~2 lines of source. Everything else is mechanical.
- **Ordering matters and is intentional:** the 403 could not honestly be called "expected" until its user-facing guidance stopped being wrong. That's why the split lands before the classification, in that order.
- **The `provider` class attribute** on the base classes shadows `ExpectedUpstreamError.provider` for the leaves via MRO. Intentional, tested.
- **Sentry rule changes are still only proposed** (Sentry is read-only for this work). Restating from #387: the `#scout-ops` rule (`10003423453`) has no filters and fires on first-seen. Even with this PR, any *new* fingerprint pages. Worth adding a `level:error` filter or swapping `FirstSeenEventCondition` for a frequency threshold.
- **Not done here:** wiring `UpstreamAccessDenied` to actually archive the membership. That's #378/#384 in wave 2 — this PR only makes the signal available and typed.

### DIFF
--- a/apps/common/errors.py
+++ b/apps/common/errors.py
@@ -63,15 +63,90 @@ class ExpectedUpstreamError(ExpectedStateError):
 # ``mcp_server/loaders/*_base.py`` — so an ``except OCSAuthError`` that imported
 # one sailed straight past the other (#371). They are defined here once and
 # imported by both: the point is identity, not a shared name.
+#
+# There are two axes, because callers need both:
+#
+#   *provider* — OCS / CommCare / Connect. Load-bearing beyond ``except``:
+#       ``_summarize_error`` serialises ``exc.__class__.__name__`` into
+#       ``MaterializationRun.result``, so the class name is what survives the
+#       JSON round-trip to the user-facing failure summary.
+#   *cause*    — 401 (the credential is dead) vs 403 (the credential is fine and
+#       has no access to this resource). #372: these need opposite advice, and a
+#       403 is an authoritative per-tenant revocation signal.
+#
+# The leaves inherit from both, so ``except OCSAuthError`` still catches every
+# OCS auth failure and ``except UpstreamAccessDenied`` catches every 403 across
+# providers.
+
+
+class UpstreamTokenExpired(Exception):
+    """HTTP 401 — the credential is dead. Reconnecting mints a working one."""
+
+
+class UpstreamAccessDenied(Exception):
+    """HTTP 403 — the credential is valid but has no access to this resource.
+
+    Distinct from a 401 because **reconnecting cannot fix it**: it mints an
+    identically-scoped credential that fails identically, so telling the user to
+    reconnect puts them in a loop (#372). The real causes are upstream access
+    removal, the resource moving to another team, or a token scoped to a
+    different team — none of which re-authenticating resolves.
+
+    Also the authoritative per-tenant revocation signal the access-revocation
+    work (#378/#384) needs: ``except UpstreamAccessDenied`` catches every 403
+    across all three providers.
+    """
+
+
+# The provider classes below are deliberately NOT expected states. They are the
+# base that ``apps/users/services/tenant_resolution.py`` raises from the login
+# signal, which today has NO user-facing surface at all — a failed resolution
+# leaves the user on an empty data-sources page and the Sentry event is the only
+# signal anything broke (#371's precondition; see rule 3 in the module
+# docstring). Classifying the base would silence that path without replacing it.
+#
+# Expectedness therefore lives on the leaf classes, which only the loaders raise.
+# The loader path DOES satisfy rule 3 — its failures reach the user through the
+# materialization failure summary in chat (apps/workspaces/tasks.py:161).
 
 
 class CommCareAuthError(Exception):
-    """Raised when CommCare HQ returns a 401 or 403."""
+    """Raised when CommCare HQ refuses our credential."""
+
+    provider = "commcare"
+
+
+class CommCareTokenExpiredError(CommCareAuthError, UpstreamTokenExpired):
+    """CommCare HQ returned 401 for a domain load."""
+
+
+class CommCareAccessDeniedError(CommCareAuthError, UpstreamAccessDenied):
+    """CommCare HQ returned 403 for a specific domain."""
 
 
 class ConnectAuthError(Exception):
-    """Raised when CommCare Connect returns a 401 or 403."""
+    """Raised when CommCare Connect refuses our credential."""
+
+    provider = "commcare_connect"
+
+
+class ConnectTokenExpiredError(ConnectAuthError, UpstreamTokenExpired):
+    """Connect returned 401 for an opportunity load."""
+
+
+class ConnectAccessDeniedError(ConnectAuthError, UpstreamAccessDenied):
+    """Connect returned 403 for a specific opportunity."""
 
 
 class OCSAuthError(Exception):
-    """Raised when Open Chat Studio returns a 401 or 403."""
+    """Raised when Open Chat Studio refuses our credential."""
+
+    provider = "ocs"
+
+
+class OCSTokenExpiredError(OCSAuthError, UpstreamTokenExpired):
+    """OCS returned 401 for an experiment load."""
+
+
+class OCSAccessDeniedError(OCSAuthError, UpstreamAccessDenied):
+    """OCS returned 403 for a specific experiment."""
--- a/apps/common/errors.py
+++ b/apps/common/errors.py
@@ -36,6 +36,8 @@ nobody is told about is not an expected state — it is a silent failure.**
 
 from __future__ import annotations
 
+from apps.common.error_codes import ErrorCode
+
 
 class ExpectedStateError(Exception):
     """A known, routine operational condition — not a defect.
@@ -49,13 +51,14 @@ class ExpectedStateError(Exception):
 class ExpectedUpstreamError(ExpectedStateError):
     """An expected state whose cause is an upstream provider, not Scout.
 
-    ``provider`` lets a handler or a log line name the system that said no
-    without re-parsing the message. Subclasses set it as a *class* attribute so
-    existing ``raise SomeAuthError("message")`` call sites keep working
-    unchanged.
+    ``provider`` and ``code`` let a handler, a log line, or a serialiser name the
+    system that said no and the condition it reported without re-parsing the
+    message. Both are *class* attributes so existing
+    ``raise SomeAuthError("message")`` call sites keep working unchanged.
     """
 
     provider: str | None = None
+    code: ErrorCode | None = None
 
 
 # These provider auth errors were each defined TWICE as unrelated classes — once in
@@ -66,10 +69,7 @@ class ExpectedUpstreamError(ExpectedStateError):
 #
 # There are two axes, because callers need both:
 #
-#   *provider* — OCS / CommCare / Connect. Load-bearing beyond ``except``:
-#       ``_summarize_error`` serialises ``exc.__class__.__name__`` into
-#       ``MaterializationRun.result``, so the class name is what survives the
-#       JSON round-trip to the user-facing failure summary.
+#   *provider* — OCS / CommCare / Connect.
 #   *cause*    — 401 (the credential is dead) vs 403 (the credential is fine and
 #       has no access to this resource). #372: these need opposite advice, and a
 #       403 is an authoritative per-tenant revocation signal.
@@ -77,6 +77,10 @@ class ExpectedUpstreamError(ExpectedStateError):
 # The leaves inherit from both, so ``except OCSAuthError`` still catches every
 # OCS auth failure and ``except UpstreamAccessDenied`` catches every 403 across
 # providers.
+#
+# The *cause* classes carry the ``code``, because that is the axis consumers
+# branch on. It is what crosses the JSON boundary into
+# ``MaterializationRun.result``; the class name is not a wire value.
 
 
 class UpstreamTokenExpired(ExpectedUpstreamError):
@@ -84,9 +88,14 @@ class UpstreamTokenExpired(ExpectedUpstreamError):
 
     Expected under the module's four-part test: known (the provider told us),
     routine (OAuth tokens expire and get revoked in normal operation), surfaced
-    (``_REAUTH_GUIDANCE`` reaches the user in the chat failure summary), and
+    (the reconnect guidance reaches the user in the chat failure summary), and
     resolved (reconnect, re-run).
+
+    Shares ``AUTH_TOKEN_EXPIRED`` with ``CredentialResolutionError``'s pre-flight
+    check: one condition gets one code however it was detected.
     """
+
+    code = ErrorCode.AUTH_TOKEN_EXPIRED
 
 
 class UpstreamAccessDenied(ExpectedUpstreamError):
@@ -107,6 +116,8 @@ class UpstreamAccessDenied(ExpectedUpstreamError):
     Classifying it before that would have silenced a condition whose only
     user-facing advice was wrong.
     """
+
+    code = ErrorCode.AUTH_ACCESS_DENIED
 
 
 # The provider classes below are deliberately NOT expected states. They are the
--- a/apps/common/errors.py
+++ b/apps/common/errors.py
@@ -79,11 +79,17 @@ class ExpectedUpstreamError(ExpectedStateError):
 # providers.
 
 
-class UpstreamTokenExpired(Exception):
-    """HTTP 401 — the credential is dead. Reconnecting mints a working one."""
+class UpstreamTokenExpired(ExpectedUpstreamError):
+    """HTTP 401 — the credential is dead. Reconnecting mints a working one.
+
+    Expected under the module's four-part test: known (the provider told us),
+    routine (OAuth tokens expire and get revoked in normal operation), surfaced
+    (``_REAUTH_GUIDANCE`` reaches the user in the chat failure summary), and
+    resolved (reconnect, re-run).
+    """
 
 
-class UpstreamAccessDenied(Exception):
+class UpstreamAccessDenied(ExpectedUpstreamError):
     """HTTP 403 — the credential is valid but has no access to this resource.
 
     Distinct from a 401 because **reconnecting cannot fix it**: it mints an
@@ -95,6 +101,11 @@ class UpstreamAccessDenied(Exception):
     Also the authoritative per-tenant revocation signal the access-revocation
     work (#378/#384) needs: ``except UpstreamAccessDenied`` catches every 403
     across all three providers.
+
+    Expected under the module's four-part test — and note rule 3 only started
+    holding for a 403 once the guidance stopped saying "reconnect" (#372).
+    Classifying it before that would have silenced a condition whose only
+    user-facing advice was wrong.
     """
 
 
--- a/mcp_server/services/materializer.py
+++ b/mcp_server/services/materializer.py
@@ -347,10 +347,17 @@ def run_pipeline(
                     }
                 raise
             except Exception as e:
-                logger.exception(
+                # WARNING, not ERROR, because this handler RE-RAISES and the
+                # caller logs the same exception at ERROR. Two logger.exception
+                # calls on one failure minted two Sentry groups every time —
+                # confirmed in production as identical pairs differing only in
+                # `logger` (#371). At WARNING this becomes a breadcrumb on the
+                # caller's single event, so the source/schema context survives.
+                logger.warning(
                     "Source %s failed for schema %s; earlier sources stay committed",
                     source.name,
                     schema_name,
+                    exc_info=True,
                 )
                 # Preserve any cursor_state advanced by per-page commits so the
                 # next run resumes from the last durable watermark (#187).
@@ -426,7 +433,13 @@ def run_pipeline(
         # Idempotent w.r.t. the per-source loop handler: if completed_at was
         # already stamped there, leave the recorded state untouched.
         if run.completed_at is None:
-            logger.exception("Materialization run %s failed before any source committed", run.id)
+            # WARNING for the same reason as the per-source handler above: this
+            # path re-raises and the caller logs the single ERROR (#371).
+            logger.warning(
+                "Materialization run %s failed before any source committed",
+                run.id,
+                exc_info=True,
+            )
             now = datetime.now(UTC)
             MaterializationRun.objects.filter(id=run.id, completed_at__isnull=True).update(
                 state=MaterializationRun.RunState.FAILED,
--- a/mcp_server/services/materializer.py
+++ b/mcp_server/services/materializer.py
@@ -50,6 +50,7 @@ from asgiref.sync import async_to_sync
 from django.utils import timezone
 from psycopg import sql as psql
 
+from apps.common.error_codes import code_of
 from apps.knowledge.services.column_note_generator import sync_column_notes
 from apps.transformations.models import TransformationAsset
 from apps.transformations.services.commcare_staging import upsert_system_assets
@@ -366,6 +367,8 @@ def run_pipeline(
                     "state": "failed",
                     "rows": (source_results.get(source.name) or {}).get("rows", 0),
                     "error": _summarize_error(e),
+                    "error_code": code_of(e),
+                    "provider": getattr(e, "provider", None) or pipeline.provider,
                     "attempts": getattr(e, "attempts", 1),
                     "failed_at": datetime.now(UTC).isoformat(),
                     "cursor_state": prior_cursor,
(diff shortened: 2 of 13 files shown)


bderenzi:

bderenzi:
Traced the "surfaced" claim end-to-end while self-reviewing, since it is what the whole classification rests on. The loader path really does reach the user in the UI, not just in agent chat text:

```
loader raises *TokenExpiredError / *AccessDeniedError
  → materializer.py:350  _summarize_error(e) → run.result["sources"][n]["error"]
  → tasks.py:128         _compose_failure_summary  (appends _REAUTH_GUIDANCE
                          and/or _ACCESS_DENIED_GUIDANCE)
  → ThreadJob.error_summary  (apps/chat/models.py:87)
  → api/jobs.ts:41
  → MaterializationFailure.tsx:67-72   rendered, with a Retry button
```

So there is a dedicated failure component displaying it — rule 3 holds on evidence rather than assertion.

The contrast with the login-signal path is total: that one has **no** equivalent surface at any layer, which is why its exception classes stay unclassified here.

Worth knowing for review: the guidance travels as a **string** through `MaterializationRun.result` (JSON), so `_looks_like_auth_failure` / `_looks_like_access_denied` have to match on text. They anchor on the exception **class name**, which `_summarize_error` puts at the front as `"{ClassName}: {message}"` — that is the stable half. Matching on message prose would have been the fragile option, and it is also why I went with subclasses rather than an enum attribute: an attribute would not have survived the JSON round-trip.

snopoke (commented):
This feels wrong. We're raising specific exception types with verbose messages but then we're also doing string parsing to check what the errors were. 

Since we control both ends of the process I think we should be able to do a much better job of generating and handling errors.

In `mcp_server/services/materializer.py` we're generating a dictionary that holds 'result' info including and error which is just a type + message. For the most part it seems we discard that error message (despite it being a verbose message) and just use it for keyword matching to determine the error category. Perhaps instead we should record and error code in the dict and then we can know exactly what the error category is without having to look for keywords in the message. If we are discarding the message then we shouldn't be writing such verbose ones.

bderenzi:
You're right, and it was worse than it looked. Fixed in 986f84a (producer) and d67e29a (consumer).

The bit that settled it: `tasks.py` was **already** writing `error_code` — at `:205` and `:330`, off `CredentialResolutionError.code`, which was added explicitly "so callers may key UI/remediation off `code`". It was read in **zero** places, while `_looks_like_auth_failure` substring-matched the same fact thirty lines up. We also already had the rule, with a finding number — `apps/agents/graph/base.py:109`:

> matched against the parsed JSON field — **NOT a substring search**: substring-matching `'"code": "NOT_FOUND"'` only worked under FastMCP's `indent=2` and would silently break under compact separators (06#1)

The materializer seam just never got it. Two more supporting details: the `transforms` half of that same `result` dict is already read as an enum comparison (`transforms.get("status") == TransformationRunStatus.FAILED`) while the `sources` half was read by substring; and the arch review already flagged `result["sources"]` as "the de-facto contract parsed independently by at least five consumers with no shared type" (2026-06-12, F-materializer).

Worth noting it isn't even a service boundary — `tasks.py:42` imports `run_pipeline`, so producer and consumer are the same process. The JSON round-trip exists only so the resume task can re-read it later.

**On "if we're discarding the message we shouldn't write verbose ones":** we're not discarding it — the resume prompt interpolates the whole summary dict into the agent's context (`tasks.py:1586`), and for the LLM verbose prose is the right interface. So the split is code-for-branching, prose-for-humans-and-the-model.

But you're onto a real defect underneath that. The messages weren't too verbose, they were doing the **wrong job**: they own remediation advice, and so does the presentation layer, so the user got it twice in two phrasings. Advice now lives in exactly one place, keyed by code. De-advising the four loader messages is the other half; it rewrites the assertions in `test_loader_401_403_split.py` that are commit 1's core pins, so I've left it for wave 1 rather than tangle it in here — the 401 case still reads its advice twice today.

**It also surfaced a bug this PR was shipping.** With a dead token on one source and revoked access on another, the two guidance blocks were appended to the end of the summary with nothing attributing either, so it rendered:

> …also failed: sessions. **This looks like an expired or revoked sign-in — reconnect the affected account** … **This is access that was removed upstream — reconnecting will NOT restore it** …

"Reconnect" and "do not reconnect", back to back, unattributed — and the denied source's message had already been dropped by the "also failed:" collapse. `test_summary_carries_both_when_two_sources_fail_differently` asserted both blocks were present and never that either named a source, so it was green on output nobody could act on. Now:

> **messages:** expired or revoked sign-in — reconnect the affected account (Settings → Connections) and re-run materialization. **sessions:** access was removed upstream, not an expired sign-in — reconnecting will NOT restore it…

Which is only cleanly possible because the code is per-source structured data rather than one substring verdict over the concatenated blob.

Two other notes: codes went into `apps/common/error_codes.py` rather than a second registry, and `mcp_server/envelope.py` now re-exports from it — one registry, and `error_response()`'s ~30 call sites are untouched. And `UpstreamTokenExpired` deliberately shares `AUTH_TOKEN_EXPIRED` with `CredentialResolutionError`'s pre-flight check, which collapses two unrelated mechanisms for "the credential is dead" into one code. Full suite 1752 passed / 5 skipped.

Picking up the rest now — de-advising the loaders, the same anti-pattern in `VIEW_SCHEMA_CASCADE_TEARDOWN_MARKER` (a sentinel string embedded in prose and substring-matched at `tasks.py:1503`), and surfacing the code to the frontend so `MaterializationFailure.tsx` can render a real "Reconnect" button instead of prose saying "Settings → Connections". One thing I'm *not* changing: `query.py:_classify_error` matching psycopg's `"password authentication failed"` stays, since we don't own that producer — string matching is legitimate as a boundary adapter, just not between two halves of our own code.

mcp_server/services/materializer.py
```diff
@@ -688,9 +692,14 @@ def _has_committed_cursor(entry: dict) -> bool:
 def _summarize_error(exc: BaseException) -> str:
     """Return a short, single-line error description for ``result["sources"][n].error``.
 
-    The string is surfaced to the agent in the resume prompt and (eventually)
-    to the end user, so it must be safe to display: no stack trace, no
-    sensitive headers, just the exception type and message.
+    The string is surfaced to the agent in the resume prompt and to the end user,
```
snopoke: Is this true - I don't see where this error get's surfaced to the agent or end user.

bderenzi: It is true, but you were right to doubt it — nothing was holding the claim up. Verified both hops and pinned them in 4309864.

**To the agent:**
```
_summarize_error → result["sources"][n]["error"]
  → _aggregate_materialization_state, detail["error"]        (tasks.py:1421-1423)
  → "Per-tenant: {summary}" in the resume prompt body        (tasks.py:1586)
  → HumanMessage(content=body) → agent.ainvoke
```

**To the end user:**
```
  → _compose_failure_summary, "{name} failed: {error}"       (tasks.py:183)
  → ThreadJob.error_summary
  → jobs/active/ recent_terminations                         (jobs_views.py:74)
  → MaterializationFailure.tsx:72, rendered at ChatMessage.tsx:286
```

The reason it was hard to confirm from the tests: `test_resume_partial_run_*` already built a run carrying `"RuntimeError: Connect 500"` and then asserted only `"completed_works" in body` and `"failed" in body` — the source *name* and the *state*, never the message. So the hop was exercised and unasserted. Both tests now assert `"Connect 500"` itself, which is what actually pins `_summarize_error`'s output rather than the dict shape around it. (The API hop was already covered by `test_jobs_endpoints.py:123`.)

One caveat I found while checking, which is probably why this smelled wrong — and it partly vindicates your original point. On the user path **only the first failed source's message survives**. `_compose_failure_summary` renders `"{first} failed ({error}); also failed: {others}"`, so with two failed sources the second one's message is discarded and it appears as a bare name. So the verbose message is genuinely dropped — just not always, and not by the mechanism it looked like. The docstring now says which hops carry it, names the test for each, and records that caveat instead of asserting an outcome.

Fixing the collapse itself is a real question but a separate one: the card would need to render a list rather than a sentence, which is a design change rather than a plumbing change. Noted for the wave-1 branch, where the card is already being touched to add the Reconnect CTA.

bderenzi:
@snopoke ready for another look — this has changed shape since your review.

**Your point had two halves and I'd only done the first.**

> If we are discarding the message then we shouldn't be writing such verbose ones.

*Record a code* (`3f8ce6f`, `af39eff`) — `apps/common/error_codes.py` is the registry, every source failure records `error_code` beside `error`, and `_looks_like_auth_failure`'s substring match is gone. You'll like the detail that settled it: `tasks.py` was **already** writing `error_code` off `CredentialResolutionError.code` at two sites, read in zero places, while substring-matching the same fact thirty lines up.

*Stop writing prose we throw away* (`69e845d`, `8cc315c`) — this is the part I'd missed, and it was in two places. All four loader raise sites carried remediation copy while the presentation layer appended its own, so a 401 told the user to reconnect **twice, in two phrasings, one sentence apart**. And `_compose_failure_summary` rendered only `failed_sources[0].error`, listing the rest as bare names — so a second failure's message was dropped entirely and read indistinguishably from a *skipped* source. That was also inconsistent with this PR's own `_credential_guidance`, which names every source the advice applies to.

The five reporting rules are now written down in `apps/common/errors.py` so a future raise site has something to violate.

**Verifying your other doubt turned up more than I expected.** You were right to ask whether `_summarize_error`'s string actually reaches anyone — it does, and both hops are now pinned by tests. But tracing it found defects on the same path that are *not* fixed here, so rather than grow this PR I wrote them up in `docs/error-contract-2026-08-11.md` (in this PR) with a three-PR plan. The two that matter:

- **A pre-flight expired token produces no guidance at all** — no `MaterializationRun` row exists yet, so none of this machinery applies. This turned out to be **#364** (`pd-wave:0`), so it goes there.
- **`access.py:66-67` and `LostAccessModal.tsx:101-103` still say "reconnect"** for an upstream access removal — #372's exact bug on two surfaces this PR doesn't touch. That's PR3 in the doc.

**#395** is already up with two more from that list (the summary being overwritten by "Please retry your request", and a 200-char cap that truncated every OCS 403 mid-word). It doesn't depend on this PR — either merge order.

Worth your eye specifically on the doc's precedent section: I checked whether this pattern already exists at Dimagi before proposing it, and `commcare-hq`'s `MessagingEvent.ERROR_MESSAGES` has been running the same code-owns-the-category / one-dict-owns-the-prose split across three surfaces for years. Scout's own MCP layer already does it too, across ~30 call sites — the Django and frontend surfaces just never adopted it.

CI is green (first real run — the old stack never matched `ci.yml`'s `branches: [main]` trigger).

apps/common/errors.py
```diff
@@ -40,26 +78,124 @@ class ExpectedStateError(Exception):
...
+#
+# The leaves inherit from both, so ``except OCSAuthError`` still catches every
+# OCS auth failure and ``except UpstreamAccessDenied`` catches every 403 across
+# providers.
+#
+# The *cause* classes carry the ``code``, because that is the axis consumers
+# branch on. It is what crosses the JSON boundary into
+# ``MaterializationRun.result``; the class name is not a wire value.
+
+
```
snopoke: unnecessary comment
```suggestion
```

apps/common/errors.py
```diff
@@ -40,26 +78,124 @@ class ExpectedStateError(Exception):
...
+    different team — none of which re-authenticating resolves.
+
+    Also the authoritative per-tenant revocation signal the access-revocation
+    work (#378/#384) needs: ``except UpstreamAccessDenied`` catches every 403
+    across all three providers.
+
+    Expected under the module's four-part test — and note rule 3 only started
+    holding for a 403 once the guidance stopped saying "reconnect" (#372).
+    Classifying it before that would have silenced a condition whose only
+    user-facing advice was wrong.
```
snopoke:
```suggestion
```

mcp_server/services/materializer.py
```diff
@@ -347,10 +348,17 @@ def _on_page(rows_loaded: int, rows_total: int | None) -> None:
                     }
                 raise
             except Exception as e:
-                logger.exception(
+                # WARNING, not ERROR, because this handler RE-RAISES and the
+                # caller logs the same exception at ERROR. Two logger.exception
+                # calls on one failure minted two Sentry groups every time —
+                # confirmed in production as identical pairs differing only in
+                # `logger` (#371). At WARNING this becomes a breadcrumb on the
+                # caller's single event, so the source/schema context survives.
```
snopoke:
```suggestion
```

mcp_server/services/materializer.py
```diff
@@ -426,7 +436,13 @@ def _on_page(rows_loaded: int, rows_total: int | None) -> None:
         # Idempotent w.r.t. the per-source loop handler: if completed_at was
         # already stamped there, leave the recorded state untouched.
         if run.completed_at is None:
-            logger.exception("Materialization run %s failed before any source committed", run.id)
+            # WARNING for the same reason as the per-source handler above: this
+            # path re-raises and the caller logs the single ERROR (#371).
```
snopoke:
```suggestion
```

bderenzi:
All four comment removals applied in `d427082`, thanks — and rebased onto current `main` while I was in there.

Two notes on how I took them:

**`errors.py:144`** — applied to the whole paragraph rather than the single line it anchored on. Deleting just that line would have left "Classifying it before that would have silenced a condition whose only" dangling mid-sentence.

**The two in `materializer.py`** explained why those handlers log at WARNING rather than ERROR. Happy to lose the prose — worth noting the rule itself is still enforced, by `tests/test_single_log_per_failure.py`, which is a static AST check over the whole module rather than a test of any one path. So a future handler that does `logger.exception` and then re-raises fails the build. That felt like the right place for it to live anyway.

Suite is **1772 passed, 5 skipped** on `main`, ruff clean.

PR #389, opened 2026-08-03
wave0(3/4): token-lifecycle error classification and the unknown-expiry hole

### LINKED ISSUE #373: Token-lifecycle error classification: expected invalid_grant logged as an exception, and unknown expiry fails open

bderenzi:
## What & why

Wave 0 of #386, PR 3 of 4. **Stacked on #388 → #387.** Closes #373.

Two small, independent defects in `apps/users/services/token_refresh.py`, one commit each.

### Defect 1 — the sync refresh logs expected failures as exceptions

`refresh_oauth_token_sync` caught bare `Exception` and called `logger.exception` for everything, including a `400 invalid_grant` — which just means the refresh token is dead, an expected outcome of normal token lifecycle. The **async twin already gets this right** (`:89-102`) and carries the reasoning in a comment; the sync (mid-run, reactive) path never received the same treatment. Mirrored it.

4xx logs at WARNING **with the response body**, so `invalid_grant` (dead token, expected) stays distinguishable from `invalid_client` (a misconfigured secret, which *is* our bug). 5xx and non-HTTP failures keep `logger.exception`.

### Defect 2 — unknown expiry fails open, and reports as healthy

`token_needs_refresh(None)` returned False — "assume valid". Since there is no validation call anywhere, `expires_at` is the **only** health signal Scout has. So a revoked-but-unexpired token was never proactively refreshed, was handed straight to the loaders, and `providers_view` fell through to `else: "connected"` — showing "Connected" in the UI right up until it 401'd. That is the mechanical link between this issue and #371's Sentry noise.

#373 asks for a decision here: fail closed, or validate against the provider. **Fail closed — but gated on there being a refresh token to act with.** That gate is the whole design:

| State | Before | After |
|---|---|---|
| unknown expiry **+** refresh token | never refreshed, served, reported "connected" | **refresh attempted.** Self-healing: success writes a real `expires_at`; failure surfaces "reconnect" up front instead of provisioning a schema and burning the discover phase on a doomed run |
| unknown expiry **+** no refresh token | served, reported "connected" | served (**unchanged**), but `providers_view` reports `"expired"` |
| known expiry | unchanged | unchanged |

Ungated fail-closed would have been a mass-reconnect risk: with nothing to refresh with, returning True condemns a credential that may well be working, for no diagnostic gain. The gate makes the null case a no-op for materialization, so the change cannot lock anyone out.

For the second row, `"expired"` is already a value the frontend renders as an amber "Reconnect" CTA (`ConnectionsPage.tsx:200-243`) — no new status value, no frontend change, and the blast radius is display-only. Asserting "Connected" on the strength of no evidence is the actual bug.

**One thing I could not verify:** how many production `SocialToken` rows have a null `expires_at` (no DB access from this environment — the tunnel wasn't up). The design deliberately doesn't depend on the answer: all three providers use django-oauth-toolkit token endpoints (`/o/token/`, `/oauth/token/`), which do return `expires_in`, and the gate neutralises the null case regardless. Happy to check before merge if you'd prefer.

|---|---|---|---|
| Sync refresh, `400 invalid_grant` (dead refresh token) | ✅ ERROR + event | 🚫 **WARNING, no event** | expected token lifecycle; the async twin already did this |
| Sync refresh, any other 4xx (401/403/429) | ✅ ERROR + event | 🚫 **WARNING, no event** | same class — provider rejected the grant |
| Sync refresh, **5xx** | ✅ ERROR + event | ✅ **unchanged** | a provider's token endpoint 500ing is not an expected state |
| Sync refresh, connection error / bug in the path | ✅ | ✅ **unchanged** | not an expected state |
| Async refresh (`refresh_oauth_token`) | 4xx already WARNING, 5xx ERROR | ✅ **unchanged** | already correct — this PR only pins it |
| `SCOUT-DJANGO-21` (`HTTPStatusError` 500 from `connect.dimagi.com/o/token/`, 9 events, culprit `/api/auth/providers/`) | ✅ | ✅ **kept** | async 5xx branch. Correct — see the caveat below |
| Token with unknown expiry served to loaders, then 401s | ✅ (as a loader `*AuthError`) | 🚫 mostly **prevented at source** | the refresh now happens before the loader runs |

**Honest caveat, the one place this can *add* events:** `providers_view` now attempts a refresh for unknown-expiry tokens that it previously skipped. If such a refresh hits a provider 500, that is a new `SCOUT-DJANGO-21` event. Group `21` is 9 events over 4 weeks, and this only affects tokens with a null `expires_at`, so the expected increase is small — but it is an increase, not a decrease, and it is real. Worth watching after deploy. The 4xx case, which is by far the more common refresh failure, is silent.

**Net on measured volume: roughly zero.** #371 recorded **zero events** for `TokenRefreshError` in 90 days, so defect 1 is preventative rather than a volume win — it closes the hole before the sync path (which only fires on a mid-run 401) starts firing. The real payoff of this PR is defect 2's correctness fix, not noise reduction. I'd rather say that plainly than claim a number.

**Pattern A — a bare `except Exception: logger.exception` wrapping an HTTP call, where a 4xx is an expected provider answer.**

- **Grep used:** `grep -rn --include="*.py" -B1 "logger.exception" apps/users/services/` and `grep -rn --include="*.py" "raise_for_status" apps mcp_server`

  - [x] `token_refresh.py:140` `refresh_oauth_token_sync` — **fixed**
  - [ ] `token_refresh.py:103` `refresh_oauth_token` (async) — **not applicable**: already branches 4xx→WARNING. It is the reference implementation, now pinned by a test asserting both functions branch the same way, since the defect was that they diverged.
  - [ ] `credential_resolver.py:142` Fernet decrypt failure — **not applicable**: a decrypt failure means key rotation or corruption, which is a genuine defect. #371 explicitly lists this class (alongside `adapters.py:48-51`) as legitimately Sentry-worthy.
  - [ ] `apps/users/providers/commcare/views.py:27` and `commcare_connect/views.py:26` — **not applicable**: `raise_for_status()` with no `logger.exception` at all. The error propagates to allauth, which renders the login-error page. No log-to-Sentry path, so nothing to reclassify.
  - [ ] `tenant_resolution.py:129, :180, :228` — **not applicable to this PR**: these raise typed `*AuthError`s rather than logging. Their messages were corrected in #388, and their classification is deliberately left alone (login path, no user-facing surface).
  - [ ] `connect_base.py:152` — **not applicable**: guarded by the typed 401/403 raise added in #388; only non-auth statuses reach it.

**Pattern B — reading token health from `expires_at` alone.**

- **Grep used:** `grep -rn --include="*.py" "token_needs_refresh\|expires_at" apps mcp_server`
  - [x] `credential_resolver.py:202` — **fixed** (passes the `can_refresh` it already computed one line above)
  - [x] `auth_views.py:280` `providers_view` — **fixed** (passes `can_refresh`, plus the honest branch for unverifiable tokens)
  - Those are the only two consumers; `token_needs_refresh` has no other callers.

## Testing

Full suite: **1763 passed, 5 skipped.**

- `tests/test_token_lifecycle_classification.py` (new, 10) — 4xx→WARNING for 400/401/403/429; 5xx and connection errors still ERROR *with* `exc_info`; the `invalid_grant` body survives into the warning; and a test pinning the sync and async implementations against each other.
- `tests/test_oauth_tokens.py` — `test_token_needs_refresh_when_none` asserted the old fail-open behaviour and is **replaced** by three tests covering both gate states and confirming `can_refresh` never overrides a *known* expiry.
- `tests/test_mcp_token_refresh.py:33` — updated call assertion; the comment records why the kwarg is there.

## Notes for reviewers

- **The judgment call is the `can_refresh` gate.** If you'd rather fail closed unconditionally, it's a one-line change — but I'd want the production null-`expires_at` count first, because that version *can* lock people out.
- **Not addressed (and out of scope for wave 0):** the deeper issue that Scout has no token *validation*, only expiry arithmetic. A `GET /me`-style liveness check per provider would make `providers_view` honest in all cases rather than inferring from `expires_at`. That is a bigger change and belongs with the revocation work in wave 2.
- `expires_at` is only written when the provider returns `expires_in` (`:111-112`, `:148-149`), so a provider that omits it leaves the old value in place — possibly stale-past, which reads as "expired" and is handled, or possibly null, which is what this PR addresses.


bderenzi:

bderenzi:
Correct, and a genuine regression in my change — fixed in 2dadc59. Verified every step of the claim before acting on it:

- `get_token_url` only knows `commcare`, `commcare_connect`, `ocs` → returns `None` for anything else → `can_refresh` is `False`.
- GitHub and Google are real configured providers: `INSTALLED_APPS` in `config/settings/base.py:78-79`, and `setup_oauth_apps.py:26` builds the `SocialApp` from `GITHUB_*`.
- `ConnectionsPage.tsx:221` gates **Disconnect** on `status === "connected"`, so the button really would have vanished.

So a GitHub-connected user would have seen a permanent amber "Connection expired", a Reconnect that can't clear it (reconnecting re-stores `expires_at=None`), and no way to disconnect — strictly worse than the bug I was fixing.

Took your suggested gate, `and token_url is not None`, with the reasoning in the comment: #373 is about credentials Scout hands to a **data loader**, and a provider with no token endpoint is a login-only identity provider whose token is never used that way.

Added two tests, both directions, so this can't silently come back:

- `test_login_only_provider_with_no_expiry_stays_connected` — GitHub with `expires_at=None`, `token_secret=""` → `"connected"`
- `test_data_provider_with_no_expiry_is_reported_expired` — OCS in the same state → `"expired"`, i.e. the #373 fail-closed behaviour is intact where it matters

Confirmed the first is non-vacuous by removing the guard and watching it fail.

Full suite at the stack tip: **1776 passed, 5 skipped**. #390 rebased onto this.

snopoke (approved):

PR #390, opened 2026-08-03
wave0(4/4): collapse four dbt Sentry fingerprints into one (662 → 138 events)

### LINKED ISSUE #374: dbt failures produce four duplicate Sentry fingerprints — 657 events, 43% of 90-day volume

bderenzi:
## What & why

Wave 0 of #386, PR 4 of 4. **Stacked on #389 → #388 → #387.** Closes #374.

The top four Scout issues by 90-day event volume are all **the same dbt failure logged four different ways** — 662 events, **43% of all 1,515 events**:

| Rank | Group | Events | Logger | What it is |
|---|---|---|---|---|
| 1 | `SCOUT-DJANGO-A` | 184 | `mcp_server.services.dbt_runner` | `dbt run failed: dbt run failed` |
| 2 | `SCOUT-DJANGO-9` | 172 | `stdout_log` | raw dbt line `Encountered an error:` |
| 3 | `SCOUT-DJANGO-F` | 168 | `file_log` | the same line, file-formatted |
| 4 | `SCOUT-DJANGO-1T` | 138 | `apps.transformations.services.executor` | `TransformStageError: Stage 'system' failed: dbt run failed` |

**Only `1T` is worth keeping** — it is the only one with a traceback, the stage name, and the underlying dbt error. The other three are strictly-less-informative copies.

### Groups 2 and 3: dbt's own loggers

Scout is **not** re-logging dbt's output — there is no stdout capture loop, no `dbtRunner(callbacks=…)`, no `logger.error(line)` anywhere in the tree. dbt fires `MainEncounteredError` and writes it to its own two real `logging.Logger`s, `stdout_log` and `file_log`, at ERROR.

Those are invisible to Django's `LOGGING` config — dbt sets `propagate = False` and clears their handlers — but **sentry-sdk patches `logging.Logger.callHandlers` directly**, so every line became an event regardless of anything in `config/settings/`. That's also why the `before_send` from #387 can't help: these records carry no exception.

Silenced with **`sentry_sdk.integrations.logging.ignore_logger`**, the SDK's supported mechanism for exactly this (thanks @snopoke — an earlier revision hand-rolled it in `before_send` off `hint["log_record"]`).

**The tradeoff, stated plainly:** `ignore_logger` disables a logger *"both in breadcrumbs and as events"* (its docstring), so dbt's raw multi-line output leaves Sentry entirely rather than being relocated onto the retained event. What survives is the error *text*, twice — the new WARNING in `dbt_runner` (a genuine breadcrumb) and the `TransformStageError` message. dbt's full output is untouched in the worker logs / CloudWatch. So the loss is the duplicate, not the diagnosis.

Considered and rejected: `--log-level none` on the dbt invocation. It would also delete dbt's output from worker logs entirely, and break `test_passes_correct_cli_args`. We want that output in CloudWatch — just not as Sentry issues.

### Group 1: the same double-log pattern as PR 2

`run_dbt` logs at ERROR and then **returns** the error; the executor re-raises it as `TransformStageError` and logs *that* at ERROR. Same rule as the materializer fix earlier in this stack: **the layer that hands the error onward logs at WARNING; the layer that raises it logs the ERROR.**

`run_dbt_test:213` looks identical and I downgraded it too — **wrongly**, caught in review. `_execute_stage` reads only `test_results["tests"]` and gates its raise on the *run* result, so nothing downstream ever inspects a test failure and no `TransformStageError` is raised for one. That ERROR is currently its only trace anywhere, so downgrading would have made dbt test failures silent rather than de-duplicated. Reverted. That the executor ignores a failing test at all is a separate correctness bug, now filed as #391.

|---|---|---|---|
| Raw dbt error line via `stdout_log` | ✅ **172 events** | 🚫 **dropped — event and breadcrumb** | dbt's own logger; the text is already in the raised exception and in the worker logs |
| Raw dbt error line via `file_log` | ✅ **168 events** | 🚫 **dropped — event and breadcrumb** | same line, second formatter |
| `run_dbt` failure (`dbt_runner`) | ✅ **184 events**, title `dbt run failed: dbt run failed` | 🚫 **WARNING, no event** | returned not raised; the executor's event is strictly better. Still a breadcrumb |
| `run_dbt_test` failure | ✅ ERROR + event | ✅ **unchanged — kept at ERROR** | no downstream logs it; it is the only trace a dbt *test* failed (#391) |
| **`TransformStageError` (`executor:91`)** | ✅ **138 events** | ✅ **kept — the one real event** | traceback + stage + underlying dbt error |
| dbt *node*-level failures (`column "x" does not exist`) | in the message | ✅ **unchanged** | still in the returned dict, the WARNING, and the raised exception |
| Any non-dbt ERROR from `mcp_server` / `apps` | ✅ | ✅ **unchanged** | the denylist is two third-party names, applied by exact logger name |
| Materializer `Transform phase failed` (`:479`) | ✅ | ✅ **unchanged** | dormant for dbt (the executor swallows), fires only on non-dbt errors |

**Net: 662 → 138 events** (all of it `run_dbt` and dbt's own loggers; no `dbt test failed` group appears in the tally, consistent with that line being left alone). A **79% reduction on this root cause, and ~35% of Scout's total 90-day Sentry volume.** Slack pings for dbt drop from up to 4 new-fingerprint pages per novel failure to 1.

### Cumulative effect of the wave

| | Events (90d) | After the stack |
|---|---|---|
| dbt family (this PR) | 662 | 138 |
| Auth family (#388) | 61 | 0 |
| Everything else | ~792 | ~792 |
| **Total** | **1,515** | **~930** |

Roughly **39% of Sentry volume removed**, and — more to the point — ranking by event count starts telling the truth. #386 warns not to trust that ranking until wave 0 lands; this is the PR that makes it trustworthy.

**Pattern — `logger.error`/`logger.exception` on a failure that is then *handed onward* (returned or re-raised) to a layer that logs it again.**

- **Grep used:** `grep -rn --include="*.py" "logger.error\|logger.exception" mcp_server/services/dbt_runner.py apps/transformations/ mcp_server/services/materializer.py`

  - [x] `mcp_server/services/dbt_runner.py:149` `run_dbt` — **fixed** → WARNING (returns the error)
  - [ ] `mcp_server/services/dbt_runner.py:213` `run_dbt_test` — **swept, then correctly excluded.** Same shape, different downstream: nothing consumes its `success`/`error`, so its ERROR is the only trace of a dbt test failure. Kept at ERROR; separate bug filed as #391. (I downgraded this initially — the sweep found the sibling, but I didn't verify it had the same downstream before applying the rule.)
  - [ ] `apps/transformations/services/executor.py:91` — **not applicable, and deliberately preserved.** This is the single event a dbt failure should produce.
  - [ ] `apps/transformations/services/executor.py:194` — **not applicable**: already `logger.warning`, correct precedent.
  - [ ] `mcp_server/services/materializer.py:479` `Transform phase failed` — **not applicable**: swallows into `transform_result`; `run_transformation_pipeline` already caught the dbt error, so this is dormant on the dbt path and is the only log on the non-dbt path.
  - [ ] `mcp_server/services/materializer.py:235, :258, :745` — **not applicable**: all swallow their exception, so each is the only record of it. (Also covered by the static check added in #388.)
  - [x] `mcp_server/services/materializer.py:351, :442` — **already fixed in #388**, same rule.

**Pattern — third-party loggers reaching Sentry through the `callHandlers` patch.** Grepped the installed dependency tree for other libraries that create named loggers and set `propagate = False`; dbt's `stdout_log`/`file_log` are the only ones currently producing events (confirmed against `sentry issue list`, where every other group's logger is a Scout module or null).

## Testing

Full suite: **1765 passed, 5 skipped.** No existing test needed editing — notably `test_passes_correct_cli_args`, which the `--log-level none` alternative would have broken.

`tests/test_dbt_sentry_noise.py` is **removed** per review: its assertions were either the SDK's own behaviour (that `ignore_logger` populates `_IGNORED_LOGGERS`) or a pin on a one-line log-level choice. The double-log rule this PR applies is already pinned generically, and statically, by `tests/test_single_log_per_failure.py` from #389.

Verified manually against the installed SDK instead: with `ignore_noisy_loggers()` applied, an ERROR on a `propagate = False` `stdout_log` produces no envelope, while `mcp_server.services.materializer` still does.

## Notes for reviewers

- **The risk to weigh is over-suppression**, and with the tests gone there is no longer an automated guard that `executor:91` stays an ERROR. It is the one event a dbt failure should produce, so it's worth a second pair of eyes on `apps/transformations/services/executor.py:91` in this diff — nothing in this PR touches it, but nothing now stops a later PR from downgrading it either.
- **The denylist is two exact third-party logger names.** `ignore_logger` fnmatches, so `stdout_log`/`file_log` cannot widen to a Scout module unless someone adds one by hand.
- **No `fingerprint` added,** though #374 floats it: `1T` already groups all 138 events into one issue with a usable title, so there is nothing for one to fix. Easy to add later if stage-level splitting turns out to be wanted.
- **Sentry rule changes still only proposed** (Sentry is read-only for this work). With volume down ~39%, the `#scout-ops` rule (`10003423453`, `actionMatch: any`, no filters, fires on first-seen) is worth revisiting — it is now the main remaining source of noise-paging rather than the event volume itself.
- **After this deploys,** the 90-day ranking in #386 should be re-derived before anyone uses it to prioritise wave 1+. That was the stated reason wave 0 goes first.


bderenzi:

bderenzi:
Both review findings were correct and are fixed in 91b47b9. Verified each against the installed SDK / the actual caller before changing anything, rather than taking them on faith:

**1. `ignore_logger` kills breadcrumbs too.** Confirmed — `sentry_sdk.integrations.logging.ignore_logger.__doc__` says it disables recording *"both in breadcrumbs and as events"*. My comment claimed the opposite.

Rather than just correcting the comment, I switched to the mechanism that actually delivers what the comment promised: drop the event in `before_send` on `hint["log_record"].name`. The `EventHandler` populates `hint["log_record"]` unconditionally, and `BreadcrumbHandler` is a separate path, so dbt's raw lines still attach to the retained `TransformStageError` event. Removed `install_logger_denylist` entirely. A test now asserts those names are **not** in the SDK's `_IGNORED_LOGGERS`, so nobody "simplifies" it back.

**2. `run_dbt_test` must keep ERROR.** Confirmed by reading `_execute_stage`: it binds `test_results`, reads only `["tests"]` per asset, and gates its raise on `result` (the *run*), never `test_results`. So a test failure raises nothing and the run is marked COMPLETED — my downgrade would have made it fully silent, which is the opposite of this PR's goal. Reverted, with the reason in a comment.

This one is my error in a specific way worth naming: the sibling sweep found the sibling, but I applied the rule to it without checking it had the same downstream. Finding the sibling is only half the rule.

That the executor ignores a failing dbt test at all is a genuine correctness bug — a `not_null` assertion failing on real NULLs reports a green run — so I filed it as **#391** rather than fixing it here. It needs a decision (fail the run vs mark PARTIAL) and is outside wave 0's observability scope.

PR body updated: the breadcrumb claim, the `run_dbt_test` row in the before/after table, and the sibling-sweep disposition. The 662 → 138 arithmetic is unaffected — as the review noted, the tally is entirely `run_dbt` and dbt's own loggers, with no `dbt test failed` group in it.

Full suite at the stack tip: **1774 passed, 5 skipped**.

config/sentry.py
```diff
@@ -18,20 +18,47 @@
...
-    classify here.
+    Everything else is kept, including a bare ``logger.error`` from Scout code
+    with no exception attached: classification is a property of an exception
+    type, and there is nothing to classify there.
     """
     exc_info = hint.get("exc_info")
     if exc_info and isinstance(exc_info[1], ExpectedStateError):
         return None
+    record = hint.get("log_record")
+    if record is not None and record.name in _NOISY_LOGGERS:
```
snopoke: Sentry has a proper mechanism for ignoring loggers: https://docs.sentry.io/platforms/python/integrations/logging/#ignoring-a-logger

tests/test_dbt_sentry_noise.py
snopoke, line 1: These tests are unnecessary

config/sentry.py
bderenzi: Done in cd02834 — switched to `ignore_logger`. That is the right call and my hand-rolled version in `before_send` (reaching into `hint["log_record"]`) was reinventing a supported API against SDK internals.

Worth flagging what changes with it, since it was the reason I moved off `ignore_logger` in the first place: it disables a logger *"both in breadcrumbs and as events"*, so dbt's raw multi-line output now leaves Sentry entirely instead of attaching to the retained `TransformStageError` event as breadcrumbs. I think that is fine and am going with it — the error *text* still reaches Sentry twice (the new WARNING in `dbt_runner`, and the `TransformStageError` message), and dbt's full output is untouched in the worker logs. The duplicate goes, not the diagnosis. Say the word if you would rather keep the breadcrumbs and I will put the `before_send` filter back.

Verified against the installed SDK rather than assuming: with `ignore_noisy_loggers()` applied, an ERROR on a `propagate = False` `stdout_log` produces no envelope, while `mcp_server.services.materializer` still does.

tests/test_dbt_sentry_noise.py
bderenzi: Removed in 1fab854.

Agreed on rereading them — they were asserting either the SDK's own behaviour (that `ignore_logger` populates `_IGNORED_LOGGERS`) or a pin on a one-line log-level choice, and the double-log rule is already covered generically and statically by `tests/test_single_log_per_failure.py` from #389.

One consequence I would rather name than bury: the last test in that file was the guard that `executor.py:91` stays an ERROR — the single event a dbt failure should produce. Nothing in this PR touches that line, but there is now no automated tripwire if a later PR downgrades it. Worth a glance during review.

Suite is 1765 passed, 5 skipped.

snopoke (approved):

